PR #32, opened 2014-08-31
added zsh file to work with oh-my-zsh

ksprashu:
oh-my-zsh needs the plugin folder name to be the same as the .plugin.zsh filename

This can be checked in oh-my-zsh.sh file in the is_plugin() function.

### DIFF
--- a/zsh-autosuggestions.plugin.zsh
+++ b/zsh-autosuggestions.plugin.zsh
@@ -1,0 +1,1 @@
+autosuggestions.plugin.zsh


faceleg:
@ksprashu does you PR impact the behaviour of this plugin under antigen in any way?

ksprashu:
@faceleg I'll check this

faceleg:
Checked?

ksprashu:
@faceleg sorry, had to get a new box. Meanwhile noticed that there is already a similar file now in the repo. So my file might be redundant.

Anyway, autosuggest is not working with either my or the tarruda repo. :-(

faceleg:
@ksprashu no problem.

Yeah it's broken. This plugin badly needs love.

PR #45, opened 2015-02-18
Added clarification to use bash

tirkarthi:
Some linux distros have sh symlinked to dash. So I have added a clarification to use bash for the distros.

### DIFF
--- a/README.mkd
+++ b/README.mkd
@@ -8,6 +8,12 @@
     $ sh ~/.zsh-autosuggestions/install
 ```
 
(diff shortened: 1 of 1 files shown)


faceleg:
Thanks!

twe4ked:
Another way to fix this could be to change the shebang to bash (or Zsh) in the install script.

faceleg:
Done: 533325c39d412b9b72fe5a23330d84d68a2ce266

PR #49, opened 2015-04-14
Simplify installation instructions

0atman:
My rational for this is:
 1. Most linux desktop users are using some kind of debian/ubuntu/mint, so why not have them catered for primarily?
 2. The bash instructions work for everyone who has bash. Last time I checked, that was 99% of the linux community, I'd say that for those who are able to install zsh, it would be even higher.


faceleg:
Agree, thanks

0atman:
No probs, great plugin :-)

PR #51, opened 2015-05-03
Install script now looks for .zshrc in $ZDOTDIR as well.

teto:
User can also supply as a first parameter the script to append to zsh-autosuggestions. 

Relates to https://github.com/tarruda/zsh-autosuggestions/issues/50


faceleg:
Thanks!

Issue #52, opened 2015-05-06
Bind right arrow to autocomplete whole line (or single word)?


calebmeyer:
I love this plugin, it's making the switch from fish to zsh easier and easier. I was happy to see it now integrates well with the syntax highlighting plugin.

However, it doesn't emulate the default behavior of fish. In fish, if I hit right arrow when a possible completion is shown, it will complete the whole thing at once.

This plugin currently completes a single character when I press the right arrow.

Is there some way to bind right arrow to complete the whole prediction?
Is there some way to bind right arrow to complete by word?

I'm running brew installed zsh 5.0.7 and the latest oh-my-zsh in iTerm2 on Mac OS X 10.10.3 (the latest yosemite) if it makes any difference.

ababol:
See https://github.com/tarruda/zsh-autosuggestions/issues/19

AUTOSUGGESTION_ACCEPT_RIGHT_ARROW=1

joaoponceleao:
Hi, I've got that at the top of my zshrc. Is there any reason why it would not work (with zgen in this case)?

jirutka:
It should work, but try to declare it *after* loading this plugin.

joaoponceleao:
Tried that. No difference. The env variable shows up properly in the terminal though.
Tried removing all options to do with HISTCONTROL, but no change.
Having said that, the up arrow works (using it along with history plugin). So this works just as well.

Also... the following behaviour is strange (may be a separate issue):
1.Type something and autocomplete with: right arrow (char by char) or, tab, or up arrow.
2.- Do not press enter, but instead delete (parts of line or whole line)
3. Autocomplete no longer works for the same specific start of the command (i.e. will work with a completely different input than that of 2).
4. Autocomplete works again after a command is entered.

alem0lars:
I'm also having the same problem. I've added `AUTOSUGGESTION_ACCEPT_RIGHT_ARROW=1` and pressing right arrow accepts just a single character.

a-m-s:
Likewise, the option has no obvious effect. I'm using vi-mode, if that makes a difference, although the default settings appear to DTRT for vi-mode widgets.

Issue #19, opened 2014-02-18
accept suggestion with RIght arrow, when at the end of line


mihaisucan:
what the title says: I would like the suggestion accepted when I press the right arrow, at the end of line.

PR #53, opened 2015-05-12
Updated readme with more configuration options.

calebmeyer:
Since #19 and #52 both asked about this, it should probably go into the readme. There may be other options which belong here, as well, but I don't know about them.


faceleg:
Thank you :)

Issue #54, opened 2015-05-15
Provide instructions for uninstallation


ikhthiandor:

baykovr:
Can't speak for the author but here's what I did:

Remove the repository folder:

     rm ~/.zsh-autosuggestions -rf

Remove the lines appended to the end of .zshrc by the install script (by hand):
   
    # Setup zsh-autosuggestions
    source /home/you/.zsh-autosuggestions/autosuggestions.zsh
    
    # Enable autosuggestions automatically
    zle-line-init() {
        zle autosuggest-start
    }

    zle -N zle-line-init

    # use ctrl+t to toggle autosuggestions(hopefully this wont be needed as
    # zsh-autosuggestions is designed to be unobtrusive)
    bindkey '^T' autosuggest-toggle

ikhthiandor:
Thanks @baykovr for responding. I also did the same, forgot to remove the directory though.

faceleg:
Does someone want to add this to the readme ?

jirutka:
I’ve described uninstallation in the readme [here](https://github.com/tarruda/zsh-autosuggestions#uninstallation).

Issue #46, opened 2015-03-04
install script emits bad characters on cygwin


apjanke:
The `install` script has a cosmetic issue when running on Cygwin/Windows 7 for me. The `\n` characters are printed literally instead of turning in to newlines, even when explicitly run under `bash`.

![install on cygwin-win7 under mintty](https://cloud.githubusercontent.com/assets/2618447/6477483/41afeba6-c1f2-11e4-841d-106b8201020e.png)

This is on Windows 7, with Cygwin updated as of about 1/25/2015.

```
$ uname -a
CYGWIN_NT-6.1 gwydion 1.7.34(0.285/5/3) 2015-02-04 12:14 x86_64 Cygwin
```

I think this is because `echo` is non-portable when used with escape sequences or options. Different shells and even different `bash`es might behave differently on different systems. Described [here](http://unix.stackexchange.com/questions/65803/why-is-printf-better-than-echo). It'll probably affect Fedora 21 and FreeBSD as well, based on a similar issue I've seen with oh-my-zsh (robbyrussell/oh-my-zsh#3658).

Looks like it could be fixed by replacing `echo` with `printf` calls.

jirutka:
I recommend to just not use this installation script and copy those few lines manually according to new instructions in the [readme](https://github.com/tarruda/zsh-autosuggestions#installation).

PR #56, opened 2015-05-18
Add alternative install/uninstall script.

baykovr:
Instead of adding multiple conf lines to .zshrc, provide them in .zsh-autosuggestionsrc and add/remove a one line reference to it in .zshrc using "source".

This also allows future configuration changes to be made to .zsh-autosugestionsrc under version control, requiring no updates/changes in user .zshrc.

Summary:
+ Adds alt-installation method.
+ Adds uninstall method ( #54 ).
+ Resolves ( #46 ).
+ Will not double install on multiple invocations like the existing install script.
+ Creates a backup of .zshrc to .zshrc.bak
+ Compatible but untested with $ZDOTDIR

Caveats
+ Requires repository be located in $HOME/.zsh-autosuggestions (as with existing install method)

### DIFF
--- a/.zsh-autosuggestionsrc
+++ b/.zsh-autosuggestionsrc
@@ -1,5 +1,5 @@
 # Setup zsh-autosuggestions
-source $DIR/autosuggestions.zsh
+source $HOME/.zsh-autosuggestions/autosuggestions.zsh
 
 # Enable autosuggestions automatically
 zle-line-init() {
--- a/.zsh-autosuggestionsrc
+++ b/.zsh-autosuggestionsrc
@@ -1,0 +1,13 @@
+# Setup zsh-autosuggestions
+source $DIR/autosuggestions.zsh
+
+# Enable autosuggestions automatically
+zle-line-init() {
+    zle autosuggest-start
+}
+
+zle -N zle-line-init
+
+# use ctrl+t to toggle autosuggestions(hopefully this wont be needed as
+# zsh-autosuggestions is designed to be unobtrusive)
+bindkey '^T' autosuggest-toggle
--- a/alt-install
+++ b/alt-install
@@ -1,0 +1,73 @@
+#!/bin/bash
+# An alternative install script for zsh-autocomplete.
+#
+DESCRIPTION="this script adds/removes a   
+.zsh-autosuggestionsrc reference to your .zshrc file."
+ZSHRC=""
+
+function usage()
+{
+        echo $0 : $DESCRIPTION
+        echo "Usage"
+        echo -e "\t" $0 " install   : attempts to install"
+        echo -e "\t" $0 " uninstall : attempts to uninstall"
+        exit 1
+}
+
+function install()
+{
+	PATH_TO_RC="source "$(pwd)"/.zsh-autosuggestionsrc"
+	# check if line exists in zshrc, add it if missing
+	if grep -Fxq "$PATH_TO_RC" "$ZSHRC"
+	then
+		echo "currently installed, see $ZSHRC" 
+	else
+		echo "Backing up $ZSHRC to $ZSHRC.bak"
+		cp "$ZSHRC" "$ZSHRC.bak"
+		echo "Adding $PATH_TO_RC to $ZSHRC"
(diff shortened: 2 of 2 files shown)


faceleg:
@ronjouch @jirutka you want to review this?

.zsh-autosuggestionsrc
```diff
@@ -0,0 +1,13 @@
...
+# Enable autosuggestions automatically
+zle-line-init() {
+    zle autosuggest-start
+}
+
+zle -N zle-line-init
+
+# use ctrl+t to toggle autosuggestions(hopefully this wont be needed as
+# zsh-autosuggestions is designed to be unobtrusive)
+bindkey '^T' autosuggest-toggle
```
ronjouch: @baykovr what about commenting this binding by default? It might override a user binding, and it's not actually needed anyway since autosuggestions is "made to work unobtrusively without it".

jirutka: Yeah, I’ve already omitted it from the basic instructions in [readme](https://github.com/tarruda/zsh-autosuggestions/blob/master/README.md); it’s mentioned only in the configuration section.

ronjouch:
@baykovr thanks, the idea seems nice. A few questions:

* Is usage of `install` / `alt-install` scripts common among zsh plugins?
* That will require documentation, can you enrich README.md as part of your pull request?
* No potential conflict with the zgen-based installation method?

jirutka:
@faceleg As I’ve already written in #60, I believe that the “installation” script to copy few lines of code to .zshrc is counterproductive, it brings more problems than it solves *. Users should know what they have in their .zshrc file. Some plugins must be loaded in particular order to not interfere with autosuggestions.

\* Really, what it’s actually trying to solve? I don’t see any real problem to solve here, only bunch of problems that such installation scripts creates.

ronjouch:
Reconsidering my comment after @jirutka 's addition. If installation was conventionally handled by zgen, supporting it would be a real win, but here the benefits are less clear.

jirutka:
By the way, these three lines:

```sh
zle-line-init() {
    zle autosuggest-start
}
zle -N zle-line-init
```

could be also contained directly in the `autosuggestions.zsh` script. Why it’s not here? Well, perhaps because it’s a configuration that should be customizable (e.g. to play well with some other plugins).

faceleg:
It works with zgen out of the box currently, if that is what you meant @ronjouch?

ronjouch:
@faceleg I'm sure it works once installed, what I was asking was whether `alt-install` and `install` scripts were a convention used by zgen and automatically invoked when installing a new plugin.

jirutka:
@ronjouch Nope, zgen, nor any other plugin manager for ZLE that I know about, doesn’t do that.

baykovr:
The aim of the alt script was to provide an uninstall feature requested in #54. 
Also, I did not want to step on any toes by providing a new/conflicting method for configuration.  So the current alt script does exactly what the existing install script does, provide some n lines into .zshrc (in an alternative format).

The method used is inspired by .zpreztorc which is configured independently from .zshrc. I think its a clean way of configuring plugin specific settings, although there are certainly some problems with it.

I suppose a healthy discussion now, is a new installer needed and if so what are the requirements for it. I'd be happy to write something everyone is happy with, if needed.

@ronjouch 
+ I'm not sure if most zsh plugins have an install method, although things such as oh-my-zsh and zprezto write directly to zshrc during setup. 
+ I can add instructions to the readme.
+ I have not tested zgen conflicts, yet.

.zsh-autosuggestionsrc
baykovr: @ronjouch sure, I just took what the existing install script does without changes. Shall I make an update?

jirutka:
@baykovr Do not get me wrong, but could you please invest your effort to improving  autosuggestions itself instead of this? Installation script just for inserting 4 lines of simple _configuration_ code into .zshrc only brings unnecessary complexity and creates more problems. There are already 5 issues related to the installation script.

Everyone who uses ZSH should be able to read and edit her/his .zshrc file.

baykovr:
@jirutka I agree.

PR #57, opened 2015-05-18
Fix and unify formatting and code-style

jirutka:


faceleg:
I :heart: unification

jirutka:
Me too! :smile_cat:

PR #60, opened 2015-05-21
Improve readme

jirutka:
I’ve replaced instructions for the install script with code snippet that should by copied. I believe that the “installation” script to copy few lines to .zshrc is counterproductive, it brings more problems than it solves. Users should know what they have in their .zshrc file. Some plugins must be loaded in particular order to not interfere with autosuggestions.

@faceleg English is not my native language, so feel free correct my sentences.


faceleg:
Thanks again!

PR #61, opened 2015-05-21
Add widget autosuggest-execute-suggestion

jirutka:
It basically means this: go to the end of line (i.e. accept the suggestion) and then hit enter.

I have binded this widget to Cmd + Enter in my terminal and it’s extremely convenient. :smile_cat:

/cc @faceleg


257:
i have bound to '^M' :) good stuff, thanks.

PR #65, opened 2015-06-05
move in history with up/down arrow keys

amdugar:
added fish like history suggestion with up and down arrow keys. 
Thoughts/Comments?


jirutka:
This functionality already provides [zsh-history-substring-search](https://github.com/zsh-users/zsh-history-substring-search) that works well with autosuggestions. What are the benefits of your implementation compared to this plugin?

amdugar:
Oh my bad, I wasn't aware of zsh-history-substring-search plugin. I'll close the pull request.

PR #76, opened 2015-10-16
Avoid bell on all keysrtokes

thorerik:
Due to using a forked version of zsh-syntax-highlighting, you also get along the bug that causes a [bell on every keystroke][1], using zsh-users/zsh-syntax-highlighting instead I avoided this bug.

[1]: https://github.com/zsh-users/zsh-syntax-highlighting/issues/200


faceleg:
Does this break anything else?

breerly:
*ping* @thorerik

thorerik:
Not that I'm aware of, it seems to work ok for me.

faceleg:
:dancer:

Issue #84, opened 2016-01-01
Autosuggestion crashes terminal window


hoodsy:
Frequently, using the right arrow to accept a suggestion will close my terminal window, regardless of whether ```AUTOSUGGESTION_ACCEPT_RIGHT_ARROW=1``` is set. Also, after running a few commands, the up-arrow history is inaccessible: 
```
_zsh_highlight_call_widget: maximum nested function level reached
❯

~
❯
autosuggest-pause:1: maximum nested function level reached
```

Removing autosuggestion removes these issues:

```bash
source ~/.zsh/zsh-autosuggestions/autosuggestions.zsh

AUTOSUGGESTION_ACCEPT_RIGHT_ARROW=1
AUTOSUGGESTION_HIGHLIGHT_COLOR="fg=5"

zle-line-init() {
    zle autosuggest-start
}
zle -N zle-line-init
```

I'm not sure the root of this issue, but it makes using this plugin essentially impossible.

ericfreese:
Looks like this is caused by sourcing ~/.zshrc twice.

```
$ zsh --version
zsh 5.2 (x86_64-apple-darwin14.5.0)
$ zsh                           # Start zsh from bash
% source ~/.zshrc               # Re-source ~/.zshrc
% sourceSegmentation fault: 11  # Type "source" and press right arrow key
```

My .zshrc:

```
source /path/to/zsh-autosuggestions/autosuggestions.zsh

zle-line-init() {
  zle autosuggest-start
}

zle -N zle-line-init
```

ahmgeek:
+1

PR #85, opened 2016-01-20
Restore standard widgets to fix segfault when re-sourcing autosuggestions.zsh

ericfreese:
Adds `$ZLE_AUTOSUGGEST_ALL_WIDGETS` array which groups `self-insert`, `magic-space`, `backward-delete-char`, and `accept-line` with other widget arrays for easy looping.

Removes all references to `autosuggest-*-orig` widgets in favor of the ['.'-prefixed names](http://zsh.sourceforge.net/Doc/Release/Zsh-Line-Editor.html#Zle-Widgets) (i.e. `.self-insert` instead of `autosuggest-self-insert-orig`)

Extracts widget setup code to two new functions: `autosuggest-hook-widgets` and `autosuggest-restore-widgets`

Calls `autosuggest-restore-widgets` when sourced to fix issue #84.

Tested locally with zsh 5.0.7 and 5.2.

### DIFF
--- a/autosuggestions.zsh
+++ b/autosuggestions.zsh
@@ -46,21 +46,23 @@ ZLE_AUTOSUGGEST_ACCEPT_WIDGETS=(
 	vi-add-next vi-forward-blank-word vi-end-of-line end-of-line
 )
 
+ZLE_AUTOSUGGEST_ALL_WIDGETS=(
+	self-insert magic-space backward-delete-char accept-line
+	$ZLE_AUTOSUGGEST_ACCEPT_WIDGETS
+	$ZLE_AUTOSUGGEST_SUSPEND_WIDGETS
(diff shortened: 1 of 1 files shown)


faceleg:
Have the installation / usage instructions changed as well? 

What I mean to say is, I want to use this IMMEDIATELY, what do I need to know?

ericfreese:
> Have the installation / usage instructions changed as well?

They shouldn't have changed. It should just work as it did before, but with less segfaults.

But you know, I'm playing with it now, and it definitely seems broken... Maybe not ready to merge just yet.

After working on ericfreese/zsh-autosuggestions though, I'm personally inclined to just go that route instead of messing with fixing this.

faceleg:
If you're saying it is more broken than it was before - I don't believe you.

Want to paste your fork on top of this?

PR #87, opened 2016-01-20
Clean up

ericfreese:
Rewriting a lot of code for improved stability.

- Using POSTDISPLAY instead of LBUFFER and RBUFFER.
- Removing pause/suspend/resume state management code.
- Removing unused server/client code.

See issue #86.

## TODO

- [ ] Remove suggestion (POSTDISPLAY) on ctrl-c. [Not sure how to do this. Ideas?]
- [x] Fix syntax highlight coloring.
- [ ] Display suggestion again after navigating history.
    - After pressing up and then down, should display the same autosuggestion it originally did.
- [x] Multi-line support
- [x] support for overriding user-defined widgets (like history-substring-search-up/down)

Issue #90, opened 2016-01-27
Cannot tab through multiple choices


wernight:
## Setup

Installed as described as OMZ plugin (so it looks like `plugins=(...  zsh-syntax-highlighting zsh-autosuggestions)`). `COMPLETION_WAITING_DOTS` is unset.

    $ bindkey "^I"
    "^I" expand-or-complete

## Bug

Doing `ls <tab>` shows a list of possible arguments (and the gray autosuggestion). Pressing <tab> again chooses the first one, pressing it again should select the second possible choice but instead completes starting from generated string:

    $ ls <tab>

    $ ls <tab>
    foo bar

    $ ls foo/<tab>
    foo bar

    $ ls foo/123<tab>
    foo bar

    $ ls foo/123/456<tab>
    foo bar

If I disable the OMZ plugins `virtualenvwrapper` and `mvn` (2 out of 12 plugins), then <tab> simply erases what comes after the cursor.

ericfreese:
There's a rewrite underway in PR #91. It's available on branch `v0.1.x`. Please try using that branch, and let me know if it solves your problem.

wernight:
Fixed on that branch.

ericfreese:
@wernight Are you still having this problem? Are you sure you're sourcing the script correctly. Can you post your .zshrc?

ericfreese:
Closing this. Please open a new issue with more specific details if you're still having trouble.

wernight:
Still having the error, here the minimal `~/.zshrc` to reproduce with my config following PR #91 (I tried some other variations as well):

````
ZSH=$HOME/.oh-my-zsh
plugins=(zsh-syntax-highlighting zsh-autosuggestions)
source $ZSH/oh-my-zsh.sh
autosuggest_start
````

Gives `command not found: autosuggest_start`. I did a git pull on `v0.1.x` just before.

wernight:
By sourcing both files it works but the issue isn't completely gone.

What works:

    $ ls xx<tab>
    $ ls xx1

What doesn't work:

    $ ls <tab>
    $ ls subfolder<CursorHere>~/Desktop

Also bugged same way:

    $ ls <Up/Down>
    $ ls temp1/a1/b1<CursorHere>~/Desktop

ericfreese:
>`plugins=(zsh-syntax-highlighting zsh-autosuggestions)`

This isn't an oh-my-zsh plugin. Please read the installation instructions in the readme.

[...]

PR #91, opened 2016-01-27
Cleanup

ericfreese:
Already pulled `ericfreese/zsh-autosuggestions#rewrite` into tarruda repo as `v0.1.x` branch so that it is available via `antigen bundle tarruda/zsh-autosuggestions --branch=v0.1.x zsh-autosuggestions.zsh`.

This PR is for merging v0.1.x into master.

See PR #87 for history.

### Fixes/changes:

- Works with zsh-syntax-highlighting plugin.
- Works with zsh-history-substring-search plugin (with a little extra config).
- Works with multi-line commands.
- Works with tab completion.
- Fixes a lot of bugs by using POSTDISPLAY instead of LBUFFER/RBUFFER and removing pause/suspend/resume state management code.
- Removes unused server/client code.

### Deprecations:

**Note that installation instructions have changed slightly**
Please check out the [readme in the v0.1.x branch](https://github.com/tarruda/zsh-autosuggestions/blob/v0.1.x/README.md) for more info.

1. No more `zle-line-init`. Just call `autosuggest_start` somewhere in your .zshrc after you source the plugin (and after you source zsh-syntax-highlighting, if you are using it):

    ```
    autosuggest_start
    ```

2. The config variables have changed:

    `AUTOSUGGESTION_HIGHLIGHT_COLOR` has become `ZSH_AUTOSUGGEST_HIGHLIGHT_STYLE `, and the other options have been removed.

Old configurations will function for now, but will display deprecation warnings.

### To use this:
- Install the v0.1.x branch with antigen
  - `antigen bundle tarruda/zsh-autosuggestions --branch=v0.1.x zsh-autosuggestions.zsh`

*OR*

- Install manually
  - `git clone -b v0.1.x https://github.com/tarruda/zsh-autosuggestions.git ~/.zsh/zsh-autosuggestions`

and source as described in the readme.

### TODO

- [x] #96
- [x] Make old `zle autosuggest-start` init and `AUTOSUGGESTION_HIGHLIGHT_COLOR` work with deprecation warnings.
- [x] `end-of-line` widget should accept the autosuggestion, regardless of where in the buffer the cursor is.
- [ ] Warn users about using zle-line-init (re: #101)

### Nice-to-haves
- [x] #95 
- [x] Make it work w/ zsh v5.0.x
- [x] Add config option to enable partially accepting autosuggestion, so `forward-char` accepts just one character, `forward-word` accepts just one word, etc. (#93)


wernight:
What about having:

````
# Enable autosuggestions
zle-line-init() {
    autosuggest_start
}

zle -N zle-line-init
````

as

````
# Enable autosuggestions
autosuggest_start
````

To avoid such breaking change in the future?

**Comment by ericfreese** [comment]
I thought about this, but ultimately decided against it. I believe the original purpose of the more verbose `zle-line-init` definition was to allow more flexibility in case you wanted to add more plugins or other functions to your `zle-line-init`. For example:

```
# Enable autosuggestions
zle-line-init() {
    another_thing_start
    autosuggest_start
    some_other_plugin_start
    ...
}

zle -N zle-line-init

hcgraf:
I'm sorry, branch v0.1.x it seems broken for me…

1. I don't get any autosuggestions, without warnings or errors
2. The cursor disappears randomly, e.g. if I type "c" it disappears, typing "d" in appears again, with "space" it disappears for a few letters, and so on…
3. The highlighting of other plugins is broken (history-substring-search again), however the syntax-highlighting plugin seems to work

BTW, for the zle-line-init, there is a plugin which provides such widgets as hooks: https://github.com/willghatch/zsh-hooks
This allows multiple plugins to extend those widgets.

ericfreese:
> I don't get any autosuggestions, without warnings or errors

> The cursor disappears randomly, e.g. if I type "c" it disappears, typing "d" in appears again, with "space" it disappears for a few letters, and so on…

> The highlighting of other plugins is broken (history-substring-search again), however the syntax-highlighting plugin seems to work

Well that's no good.

Can you open a separate issue with more information:

- what does your .zshrc look like?
- what other plugins are you using?
- what version of zsh?
- what operating system?
- a screenshot or gif/movie showing the problem would be great

As far as hooking built-in widgets goes, I did end up moving away from aliasing the dot-prefixed ones. It is a little bit less stable, but there was an issue getting autocompletions to work that was easily fixed by doing so.

faceleg:
Wow this looks awesome

george-b:
I have been using the v0.1.x branch for the last week or so without issue :+1:

faceleg:
@ericfreese can we merge?

wernight:
I have some issues that I reported. It's better but not yet great.

For one, is the `zsh-autosuggestions.plugin.zsh` used to work and now doesn't. Another issue is reported in #90.

faceleg:
So #90 isn't fixed?

wernight:
@faceleg No, see [my comment](https://github.com/tarruda/zsh-autosuggestions/issues/90#issuecomment-177920643), there is too much auto-complete and it's a very frequent use case.

ericfreese:
I'm going to go ahead and merge this tonight.

sorin-ionescu:
In order to integrate `zsh-autosuggestions` into Prezto, I have to added the following to `modules/autosuggestions/init.zsh`. 

```zsh
if zstyle -t ':prezto:module:history-substring-search' loaded; then
  ZSH_AUTOSUGGEST_CLEAR_WIDGETS+=(
    history-substring-search-up
    history-substring-search-down
  )
fi
```

In my testing, I have observed that the `if statement` is unnecessary because adding those widgets to the array, even when `zsh-history-substring-search` is not sourced, causes no errors. So, why not add the widgets directly to https://github.com/zsh-users/zsh-autosuggestions/blob/master/src/config.zsh#L15 and excuse the user from having to do it?

ericfreese:
> why not add the widgets directly

The idea was to avoid as much as possible supporting other plugins inside of this plugin, since it generally leads to bloat in the code. In this case though, I've gone ahead and added those widgets to the default configuration in `v0.2.17`, since the change is pretty minimal, and like you said, it doesn't cause any errors.

sorin-ionescu:
The following should be enough to integrate zsh-autosuggestions into Prezto.

```zsh
pmodload 'editor'

source "${0:h}/external/zsh-autosuggestions.zsh" || return 1

zstyle -s ':prezto:module:autosuggestions:color' found \
  'ZSH_AUTOSUGGEST_HIGHLIGHT_STYLE' || ZSH_AUTOSUGGEST_HIGHLIGHT_STYLE='fg=8'

if ! zstyle -t ':prezto:module:autosuggestions' color; then
  ZSH_AUTOSUGGEST_HIGHLIGHT_STYLE=''
fi

if [[ -n "$key_info" ]]; then
  bindkey -M viins "$key_info[Control]F" vi-forward-word
  bindkey -M viins "$key_info[Control]E" vi-add-eol
fi
```

PR #122, opened 2016-02-21
Add an autosuggest widget: autosuggest-execute.

hitripod:
Add back `autosuggest-execute-suggestion` from #61 which is removed after rewriting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,7 @@ This plugin works by triggering custom behavior when certain [zle widgets](http:
 
 - `ZSH_AUTOSUGGEST_CLEAR_WIDGETS`: Widgets in this array will clear the suggestion when invoked.
 - `ZSH_AUTOSUGGEST_ACCEPT_WIDGETS`: Widgets in this array will accept the suggestion when invoked.
+- `ZSH_AUTOSUGGEST_EXECUTE_WIDGETS`: Widgets in this array will execute the suggestion when invoked.
 - `ZSH_AUTOSUGGEST_PARTIAL_ACCEPT_WIDGETS`: Widgets in this array will partially accept the suggestion when invoked.
 
 Widgets not in any of these lists will update the suggestion when invoked.
@@ -78,10 +79,11 @@ Widgets not in any of these lists will update the suggestion when invoked.
 
 ### Key Bindings
 
-This plugin provides two widgets that you can use with `bindkey`:
+This plugin provides three widgets that you can use with `bindkey`:
 
 1. `autosuggest-accept`: Accepts the current suggestion.
-2. `autosuggest-clear`: Clears the current suggestion.
+2. `autosuggest-execute`: Accepts and executes the current suggestion.
+3. `autosuggest-clear`: Clears the current suggestion.
 
 For example, this would bind <kbd>ctrl</kbd> + <kbd>space</kbd> to accept the current suggestion.
 
--- a/src/config.zsh
+++ b/src/config.zsh
@@ -30,6 +30,10 @@ ZSH_AUTOSUGGEST_ACCEPT_WIDGETS=(
 	vi-end-of-line
 )
 
+# Widgets that accept the entire suggestion and execute it
+ZSH_AUTOSUGGEST_EXECUTE_WIDGETS=(
+)
+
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -81,6 +82,7 @@ Widgets not in any of these lists will update the suggestion when invoked.
 This plugin provides two widgets that you can use with `bindkey`:
 
 1. `autosuggest-accept`: Accepts the current suggestion.
+2. `autosuggest-execute`: Accepts the current suggestion and execute it.
```
ericfreese: Let's reword this. How about "Accepts and executes the current suggestion."?

src/config.zsh
```diff
@@ -30,6 +30,14 @@ ZSH_AUTOSUGGEST_ACCEPT_WIDGETS=(
 	vi-end-of-line
 )
 
+# Widgets that accept the entire suggestion and execute it
+ZSH_AUTOSUGGEST_EXECUTE_WIDGETS=(
```
ericfreese: I think this array should be empty by default. All of these widgets already exist in the ACCEPT array, and shouldn't be in both. Users who want these widgets to execute instead of accept could then remove these widgets from the ACCEPT array and add to the EXECUTE array in their own configuration after sourcing the plugin

src/widgets.zsh
```diff
@@ -47,6 +47,23 @@ _zsh_autosuggest_accept() {
 	_zsh_autosuggest_invoke_original_widget $@
 }
 
+# Accept the entire suggestion and execute it
+_zsh_autosuggest_execute() {
+	# Only accept if the cursor is at the end of the buffer
+	if [ $CURSOR -eq $#BUFFER ]; then
```
ericfreese: It looks like the original implementation didn't have this check to see if the cursor is at the end of the line. I think we should go ahead and remove this if statement and execute the suggestion regardless of where the cursor is.

src/widgets.zsh
```diff
@@ -47,6 +47,23 @@ _zsh_autosuggest_accept() {
...
+		# Add the suggestion to the buffer
+		BUFFER="$BUFFER$POSTDISPLAY"
+
+		# Remove the suggestion
+		unset POSTDISPLAY
+
+		# Move the cursor to the end of the buffer
+		CURSOR=${#BUFFER}
```
ericfreese: Moving the cursor shouldn't be necessary when executing the suggestion

src/widgets.zsh
```diff
@@ -47,6 +47,23 @@ _zsh_autosuggest_accept() {
...
+	fi
+	_zsh_autosuggest_invoke_original_widget $@
+	_zsh_autosuggest_invoke_original_widget accept-line
```
ericfreese: I think we should be able to just go ahead and use `zle .accept-line` here instead

hitripod:
@ericfreese Thanks for your nice review! Revised.

src/widgets.zsh
```diff
@@ -47,6 +47,11 @@ _zsh_autosuggest_accept() {
 	_zsh_autosuggest_invoke_original_widget $@
 }
 
+# Accept the entire suggestion and execute it
+_zsh_autosuggest_execute() {
```
ericfreese: I believe we still need to [set the buffer](https://github.com/tarruda/zsh-autosuggestions/pull/122#discussion-diff-53573072R55) and [clear the postdisplay](https://github.com/tarruda/zsh-autosuggestions/pull/122#discussion-diff-53573072R58) here before accepting the line.

hitripod: Yes, you're right. It's my mistake.

README.md
```diff
@@ -81,7 +82,8 @@ Widgets not in any of these lists will update the suggestion when invoked.
 This plugin provides two widgets that you can use with `bindkey`:
```
ericfreese: Can you update this line to say "...provides three widgets..."?

ericfreese:
Can you update your PR to merge into the [`features/execute_widget`](https://github.com/tarruda/zsh-autosuggestions/tree/features/execute_widget) branch?

hitripod:
Close to retarget the branch. See PR #124.

PR #124, opened 2016-02-22
Add an autosuggest widget: autosuggest-execute.

hitripod:
(Originally PR #122, retarget it to features/execute_widget branch.)

Add back autosuggest-execute-suggestion from #61 which is removed after rewriting.


ericfreese:
Thanks @hitripod. I want to play with this a little bit myself before releasing, and will probably release this with v0.3.0. Would be great if we could add tests as well. Take a look at [`scripts/test.zsh`](https://github.com/tarruda/zsh-autosuggestions/blob/master/script/test.zsh) for examples of how we're testing the other widgets, and send another PR if/when you get some tests in. Otherwise, I'll probably write some tests for this new widget at some point before releasing v0.3.0.

hitripod:
Thank you, too. I would love to do that :)

Cottser:
Just wanted to say thanks for this! I was looking for this functionality (just started using zsh-autosuggestions) and appreciate that this is back. I've added the widget code with a different function name to my ~/.zshrc temporarily until v0.3.0 is released.

PR #129, opened 2016-02-28
Add option for matching on previous command.

gezalore:
New feature:
Setting ZSH_AUTOSUGGEST_MATCH_PREV_CMD will enable the plug-in to be a
bit more context aware when generating the suggestion, by matching the
previously executed command against the command executed before the
preferred suggestion. See src/config.zsh for an example.

Also added testing for the suggestion computation.

### DIFF
--- a/script/test.zsh
+++ b/script/test.zsh
@@ -3,11 +3,150 @@
 SCRIPT_DIR=$(dirname "$0")
 TEST_DIR=$SCRIPT_DIR/../test
 DIST_DIR=$SCRIPT_DIR/../
+TMPHIST_FILE=/tmp/zsh-autosuggestions-test-tmp-hist
 
 # Use stub.sh for stubbing/mocking
 source $TEST_DIR/stub-1.0.2.sh
 
 source $DIST_DIR/zsh-autosuggestions.zsh
+
+#--------------------------------------------------------------------#
+# Suggestions                                                        #
+#--------------------------------------------------------------------#
+
+testSuggestionSimple() {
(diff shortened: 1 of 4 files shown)


ericfreese:
In anticipation of [many differing ideas](https://github.com/zsh-users/zsh-autosuggestions/issues?q=is%3Aissue+is%3Aopen+label%3Asuggestion-engine) of what would be the best way to fetch a suggestion, I've implemented a strategy pattern for fetching suggestions.

See branch [`features/suggestion_strategies`](https://github.com/zsh-users/zsh-autosuggestions/tree/features/suggestion_strategies) and in particular commit https://github.com/zsh-users/zsh-autosuggestions/commit/0a526667b25ba7c6dc69ab1d452d421818ee2562 that adds the [`src/strategies/`](https://github.com/zsh-users/zsh-autosuggestions/tree/features/suggestion_strategies/src/strategies) directory and the ["default" strategy](https://github.com/zsh-users/zsh-autosuggestions/blob/features/suggestion_strategies/src/strategies/default.zsh) that just picks the most recent history match. A different strategy can be selected by setting the `ZSH_AUTOSUGGEST_STRATEGY` variable (but currently the only available strategy is "default").

Can you update your code to use this new strategy pattern, adding a new strategy file to the `src/strategies/` directory? And then re-submit the PR to the `features/suggestion_strategies` branch instead of `master`?

Thanks!

gezalore:
Done, see #131. You can close this now if you feel that is appropriate. Thank you!

PR #131, opened 2016-03-01
Add match_prev_cmd strategy and fix default strategy

gezalore:
Re-raised #129 as requested. There is also a bugfix for the default strategy. Feedback is welcome.


ericfreese:
Thanks for this. Looks like good stuff! Will dig in a bit deeper as I have time over the next few days.

PR #134, opened 2016-03-12
Implement asynchronous suggestions

PythonNut:
Hey, just throwing this out there. If you guys don't want to merge it, I'm totally okay with maintaining my own fork. 

This is based on the fantastically splendid [`zsh-async`](https://github.com/mafredri/zsh-async) library. Since I don't really know how dependency management works in ZSH, I just assume that the file has been `source`d prior to this one. 

I wrote this because I wanted autosuggestions on an old Raspberry Pi, but the three-hundred millisecond pauses were making typing unpleasant. Now the suggestions don't interfere with typing. :)

Unfortunately, I couldn't figure out how to pass the `history` array to the async job efficiently, so I have to resort to reading the `$HISTFILE` to get the history. This isn't terribly efficient, but it happens in the async job, so the perceived lag is not affected. However, it could lead to some subtle errors if `[[ -o extended_history ]]` doesn't reflect the current structure of the `$HISTFILE` (not a problem for me personally because I always have it on). If you have any idea how to improve this, please tell.

### DIFF
--- a/zsh-autosuggestions.zsh
+++ b/zsh-autosuggestions.zsh
@@ -205,6 +205,57 @@ _zsh_autosuggest_highlight_apply() {
 }
 
 #--------------------------------------------------------------------#
+# Suggestion                                                         #
+#--------------------------------------------------------------------#
+
+# Get a suggestion from history that matches a given prefix
+_zsh_autosuggest_suggestion() {
+	setopt localoptions EXTENDED_GLOB
+
+	# Escape special chars in the string (requires EXTENDED_GLOB)
+	local prefix="${@//(#m)[\\()\[\]|*?]/\\$MATCH}"
+	if [[ -o extended_history ]]; then
+		local -a recent=(${"${(f)$(cat $HISTFILE)}"#*;})
+	else
+		local -a recent=(${"${(f)$(cat $HISTFILE)}"})
+	fi
+	# Echo the first item that matches
+	echo -E ${recent[(R)$prefix*]}
+}
+
+_zle_synchronize_postdisplay() {
+	if [[ -n "$POSTDISPLAY_INTERNAL" && -n $BUFFER ]]; then
+		if [[ $POSTDISPLAY_INTERNAL == "$BUFFER"* ]]; then
+			POSTDISPLAY=${POSTDISPLAY_INTERNAL#$BUFFER}
+		else
+			# The POSTDISPLAY is out-of-date so recompute it
+			unset POSTDISPLAY
+			async_job suggest _zsh_autosuggest_suggestion $BUFFER
+		fi
+	else
+		unset POSTDISPLAY
+	fi
+
+	# Force the highlighting to take place immediately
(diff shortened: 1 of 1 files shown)


tarruda:
In the repository history there's a completion server which I used to try implementing async completion at the start of the project(not only for history, but it also used the user's custom completion which could be slow). I remember it didn't work properly because of zsh limitations at the time, but maybe you can reuse something there.

PythonNut:
@tarruda I did take a look at that a while ago, but I honestly couldn't tell what was going on. (I don't blame you, `zsh` is an especially baroque language; I even have trouble reading my own code from time to time.) I think it's interesting, and ultimately I would like to revisit the idea of suggesting completions. However, I do think that `zsh-async` allows for a cleaner, easier to understand implementation.

@ericfreese you seem to be running the show now. Does this interest you? If it does, I'll try to keep this synchronized and we can work on merging it, but if not, then I'll close it.

ericfreese:
It is interesting but I'll have to dig in a bit before I can say whether
I'd want to merge or not. I'd say leave the PR alone for now (don't worry
too much about keeping it in sync), and I'll do my best to take a closer
look as soon as possible.

On Monday, March 14, 2016, PythonNut <notifications@github.com> wrote:

> @tarruda <https://github.com/tarruda> I did take a look at that a while
> ago, but I honestly couldn't tell what was going on. (I don't blame you,
> zsh is an especially baroque language; I even have trouble reading my own
> code from time to time.) I think it's interesting, and ultimately I would
> like to revisit the idea of suggesting completions. However, I do think
> that zsh-async allows for a cleaner, easier to understand implementation.
>
> @ericfreese <https://github.com/ericfreese> you seem to be running the
> show now. Does this interest you? If it does, I'll try to keep this
> synchronized and we can work on merging it, but if not, then I'll close it.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub:
>
> https://github.com/zsh-users/zsh-autosuggestions/pull/134#issuecomment-196640099
>


-- 
Sent from my iPhone

PythonNut:
@ericfreese thanks. I look forward to your feedback.

ericfreese:
@PythonNut 

Thank you very much for the work and the PR, but I don't think I want to merge this because of the added dependency on the zsh-async project. I think we can come up with a more lightweight solution. It looks like zsh-async uses pseudoterminals via the zpty module. We could use zpty directly to implement the suggestion fetching. I've also been looking a bit into using a `coproc`, but there are a few kinks I still need to work through.

Making z-asug asynchronous is high priority on my list since I think it will fix many of the existing performance issues. I just don't think this PR is quite the right way to do it.

Thanks again and sorry for such a slow response!

PythonNut:
Hey, thanks for getting back to me. If you can use the extra information specific to this problem to create a more optimal solution, then that's great! 

I only used `zsh-async` because it's easy, and I've had a lot of frustration with asynchronous tasks in `zsh` in the past. 

I look forward to a superior solution.

PR #147, opened 2016-04-20
add # symbol to escape characters

seletskiy:


danielshahaf:
In recent zsh you may be able to use ${(b)parameter} for this:

```zsh
% s='foo#bar'
% print -r - ${(b)s}
foo\#bar
```

Reference: http://zsh.sourceforge.net/Doc/Release/Expansion.html#Parameter-Expansion-Flags

ericfreese:
Added commit 5237203 to address this using the `b` param expansion flag. Thanks @danielshahaf for the heads up

ericfreese:
@danielshahaf Do you know when the `b` param expansion flag was added to zsh?

seletskiy:
@ericfreese: And how this is going to help? I'm using `#` in my command line as an alias for various commands: https://github.com/seletskiy/zsh-hash-aliases

danielshahaf:
> @danielshahaf Do you know when the `b` param expansion flag was added to zsh?

Commit zsh-users/zsh@6269db883ad6989a578bac299deab9c4285bb1df, first released in 5.1.1, workers/35067.

ericfreese:
@danielshahaf Looking into it a bit more, it looks like it was actually first released in 5.0.8: https://github.com/zsh-users/zsh/blob/zsh-5.0.8/ChangeLog#L314-L319

@seletskiy This would help by escaping `#` characters. I believe the `b` flag should escape all chars [here](https://github.com/zsh-users/zsh/blob/zsh-5.0.8/Src/zsh.h#L220).

--

Unfortunately, I think we should try to support zsh versions below 5.0.8, so will probably revert the above commit and keep the escape function, adding a note that we may be able to use the `b` flag at some point in the future.

ericfreese:
@seletskiy I'm also not seeing an issue to begin with here. `#` characters seem to work just fine for me with the code as-is (tested w/ versions 5.0.8 and 5.2.

```shell
% zsh -f
%% alias -g -- '#wc'='| wc -l'
%% source zsh-autosuggestions.zsh
%% cat zsh-autosuggestions.zsh #wc
     429
%% cat zsh-autosuggestions.zsh #       <== Suggestions work fine here
```

Can you explain more specifically the exact problem you're having? What version of zsh are you using?

danielshahaf:
@ericfreese I stand corrected.

seletskiy:
Please, reopen. It is the issue, otherwise I haven't reported it.

```
 λ  ~ alias -g -- '#'='| grep'
 λ  ~ echo 1 # 1
1
 λ  ~ _zsh_autosuggest_strategy_match_prev_cmd:18: bad math expression: operator expected at `1 # 1'
#_zsh_autosuggest_strategy_match_prev_cmd:18: bad math expression: operator expected at `1 # 1'
#
```

Alok:
I'm also getting this issue. Typying `#` gives me the same error.

PR #148, opened 2016-04-22
suspend suggestions on fast changes

seletskiy:
Because right now insert is slow, arrow movements are slow.

### DIFF
--- a/src/config.zsh
+++ b/src/config.zsh
@@ -1,3 +1,4 @@
+zmodload zsh/datetime
 
 #--------------------------------------------------------------------#
 # Global Configuration Variables                                     #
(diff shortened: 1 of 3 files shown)


ericfreese:
While I like the idea behind this PR, I don't like some of the unintended consequences.

The biggest problem I see is that if you type fast enough, you won't get a suggestion at the end of your typing. For example, say you have `echo hello` in your history and type the characters `e` and `c` in quick succession (faster than the threshold of 0.05 seconds). You won't get a suggestion until you type `h`. IMO that's a deal breaker for this PR.

What we really want here is a debouncing mechanism, but I'm not sure the best way to achieve it in zsh. #134 adds async support and may address this, but I don't like the added dependency on `zsh-async`. Might be worth looking into how `zsh-async` works and/or shooting off an email to one of the zsh email lists.

seletskiy:
@ericfreese: if you type fast as I do, you will likely to type far ahead of autocompletion speed. Also, if you still want to see autocompletion, you just type/erase one char and see completion. Also, `0.05` is not hardcoded and can be configured by user.

Actually, current slowdown is a dealbreaker for using that plugin whatsoever for fast typists.

ericfreese:
> if you still want to see autocompletion, you just type/erase one char and see completion

Sorry, but I don't want to introduce this behavior. I think we can find another solution for fast typists that keeps the behavior of the plugin as-is.

Issue #118, opened 2016-02-15
Interaction with completion menu


kergoth:
When it inserts the current choice from the completion menu, the grey autosuggest sticks around and changes to the normal color. See https://asciinema.org/a/ateiyie18acvm3l7srfezt89g for a screencast of the behavior. bindkey '^I' is expand-or-complete.

ericfreese:
This is what I was referring to as "minor issues with the buffer displaying incorrectly while inside menu selection" in this [comment](https://github.com/tarruda/zsh-autosuggestions/issues/73#issuecomment-176504444), but I never created a separate issue for the problem.

Using menu selection is what causes this issue. I believe this is the minimal `.zshrc` config to reproduce this problem:

```
autoload compinit && compinit
zstyle ':completion:*' menu select
source ~/.zsh/zsh-autosuggestions/zsh-autosuggestions.zsh
```

I haven't yet found a way to bind to a menu selection event to clear/update the suggestion. Probably will need to dig into [this document](http://zsh.sourceforge.net/Doc/Release/Completion-System.html) to find a solution.

As far as I know, there are two options to workaround this issue temporarily:

1. Don't use menu selection. Remove the zstyle line above or set the menu style to `no-select` with `
zstyle ':completion:*' menu no-select`. This will keep suggestions working while doing completions.
2. Add your bound completion widget(s) (`expand-or-complete` in your case) to [the list of widgets that clear the suggestion](https://github.com/tarruda/zsh-autosuggestions/blob/master/src/config.zsh#L15-L23). This will have the effect of disabling suggestions while you're in menu selection, but at least it won't leave a lingering, broken suggestion. Add `ZSH_AUTOSUGGEST_CLEAR_WIDGETS+=("expand-or-complete")` somewhere in your `.zshrc` after sourcing this plugin.

We may want to add `expand-or-complete` to `ZSH_AUTOSUGGEST_CLEAR_WIDGETS` by default, but I believe that would break suggestions for folks not using menu selection, so I think this belongs in individual user configuration for now.

kergoth:
Ah! Thanks for the detailed and quick response, much appreciated.

ericfreese:
This should be fixed in v0.3.2 (just released) by commit e87bc74

gustavoruiz24:
I had the same problem, adding `zinit ice wait lucid atload zsh-autocomplete` solved it. Here is part of my .zshrc:

```
# Plugins
zinit ice depth=1
[...]

PR #149, opened 2016-04-22
Remove suggestion on complete

seletskiy:


ericfreese:
What problem is this solving?

seletskiy:
Before:
zsh-sug-before

After:
zsh-sug-after

ericfreese:
Ok. It sounds like you're describing #118. I think this is a good solution. Commit e87bc74 should implement the same behavior, using `unset` instead of setting it to an empty string and including a comment.

Thanks!

PR #161, opened 2016-05-28
Document caveats of match_prev_cmd strategy

tsdh:
This strategy relies on the history being exactly in the order in which
commands have been entered.  Therefore, options like suppressing
duplicates or expiring duplicates first will lead to unexpected
suggestions.


ericfreese:
Thanks for the PR! I've just changed the branching scheme so all PRs should come in to the `develop` branch instead of `master`. Can you please re-submit your PR to the `develop` branch?

tsdh:
Done!

PR #163, opened 2016-05-28
Document caveats of match_prev_cmd strategy

tsdh:
This strategy relies on the history being exactly in the order in which
commands have been entered.  Therefore, options like suppressing
duplicates or expiring duplicates first will lead to unexpected
suggestions.


ericfreese:
Thanks for this. Small thing but can you run 'make' and add the resulting change to your commit?

tsdh:
Yes, of course. Done!

ericfreese:
:+1:

PR #177, opened 2016-07-15
Disable autosuggest if buffer is too large

lbolla:
This is a workaround when copying large amount of texts in a terminal, i.e. with mouse-button-2.
The idea is to disable autosuggestion when buffer is too large. This avoids that the autosuggest function is called too many times (it's called once per character), slowing it down the paste.
I hardcoded the max buffer length to 20 chars. Possibly, it could be a configurable value instead.


lbolla:
Submitted new PR https://github.com/zsh-users/zsh-autosuggestions/pull/178.

PR #178, opened 2016-07-18
Disable autosuggest if buffer is too large

lbolla:
Make buffer max size configurable, defaulted to infinity ($ZSH_BUFFER_MAX_SIZE).

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,12 @@ Widgets not in any of these lists will update the suggestion when invoked.
 **Note:** A widget shouldn't belong to more than one of the above arrays.
 
 
+### Disabling suggestion for large buffers
+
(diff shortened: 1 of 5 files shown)


z0rc:
Please document the new configuration variable. Also this variable should be under `ZSH_AUTOSUGGEST_` "namespace" for consistency with other configuration variables.

z0rc:
Can you also update https://github.com/zsh-users/zsh-autosuggestions/blob/develop/README.md#configuration with new configuration value? That will make sure that the regular user knows about this option without checking the source code.

z0rc:
Thank you very much! @ericfreese gentle ping.

ericfreese:
Thank you @lbolla and @z0rc for your work here!

Issue #170, opened 2016-06-10
Fetch Suggestions Asynchronously


ericfreese:
Will solve a number of performance problems.

Was implemented by @PythonNut using `zsh-async` in PR #134, but I think we should be able to implement it without adding the dependency.

Seems to me the only real way to make this asynchronous is to have another process that's responsible for fetching the suggestion. `zsh-async` uses `zpty`s, which might be the best solution. I've also looked a bit into `coproc` (see gist [here](https://gist.github.com/ericfreese/20db47bc62b9df3e28732ea03c1bc811)), but mostly decided they're not the best way forward.

Next will be looking into `zpty`.

PythonNut:
:+1: looking forward to a better solution.

ericfreese:
Merged into `develop`.

PR #180, opened 2016-07-20
Asynchronous suggestions

ericfreese:
Uses [zpty](http://zsh.sourceforge.net/Doc/Release/Zsh-Modules.html#The-zsh_002fzpty-Module) to fetch suggestions asynchronously. 

Work in progress. Based on mafredri's [zsh-async](https://github.com/mafredri/zsh-async).

See issue #170 and previous PR #134.
## 
### TODO
- [x] Support older zsh versions (5.0.x) where `zpty` doesn't give a file descriptor.
- [x] Support both sync and async modes via config option for environments where `zpty` doesn't work (see https://github.com/sindresorhus/pure/issues/117)
- [x] Update/rewrite broken tests.
- [x] Fix [suggestion appearing intermittently in `mc`](https://github.com/zsh-users/zsh-autosuggestions/pull/180#issuecomment-275875894)
- [x] Fix [suggestions failing to become invalidated](https://github.com/zsh-users/zsh-autosuggestions/pull/180#issuecomment-275911411)
- [x] Fix  [stderr output showing up in suggestions](https://github.com/zsh-users/zsh-autosuggestions/pull/180#issuecomment-276071054).
- [x] Look into [processes running in `zpty` receiving `SIGHUP` when any `zpty` is deleted](https://github.com/zsh-users/zsh-autosuggestions/pull/180#issuecomment-276071054).
- [x] Find workaround for `zpty -d` killing all previously created `zpty` commands (or if no workaround, default to synchronous suggestions for now).


balta2ar:
@ericfreese ping?

ericfreese:
@balta2ar Still alive, but I've been spending a lot of time on other projects lately. Still intending to get to this when I have more time.

z0rc:
Ping?

sindresorhus:
// @mafredri

src/widgets.zsh
```diff
@@ -33,18 +33,12 @@ _zsh_autosuggest_modify() {
 	fi
 
 	# Get a new suggestion if the buffer is not empty after modification
-	local suggestion
 	if [ $#BUFFER -gt 0 ]; then
 		if [ -z "$ZSH_AUTOSUGGEST_BUFFER_MAX_SIZE" -o $#BUFFER -lt "$ZSH_AUTOSUGGEST_BUFFER_MAX_SIZE" ]; then
-			suggestion="$(_zsh_autosuggest_suggestion "$BUFFER")"
+			_zsh_autosuggest_async_fetch_suggestion "$BUFFER"
```
mafredri: Might I suggest [feature detection](https://github.com/mafredri/zsh-async/blob/a4b2f81c966a00eeb38876d505d97423cc8addfd/async.zsh#L395-L401)? Takes care of supporting both sync/async, keeps the code simple and doesn't require a config option.

src/async.zsh
```diff
@@ -0,0 +1,53 @@
...
+		print -n -- "$(eval "$cmd")"$'\0' &
+
+		# Save the bg process's id so we can kill later
+		last_pid=$!
+	done
+}
+
+_zsh_autosuggest_async_suggestion_ready() {
+	# while zpty -rt $ZSH_AUTOSUGGEST_PTY_NAME suggestion 2>/dev/null; do
+	while read -u $_ZSH_AUTOSUGGEST_PTY_FD -d $'\0' suggestion; do
```
mafredri: Using `zpty -r` to read the FD here would allow supporting older versions of zsh e.g. through `kill`-signals. Doesn't support delimiter, but you can use a PATTERN as the last parameter, e.g. `zpty -rt $ZSH_AUTOSUGGEST_PTY_NAME suggestion '*'$'\0'`. I've not used pattern before, though, so I could be wrong on the syntax.

z0rc:
> Is there any particular reason for not using zsh-async here, btw?
See https://github.com/zsh-users/zsh-autosuggestions/pull/134#issuecomment-224076525
Personally I believe it's okay to depend on zsh-async though.

mafredri:
Thanks @z0rc. I think it makes sense seeing as z-asug has quite simple requirements. Benefits of `zsh-async` only start to show when you require more advanced job control (or need to support older versions of zsh).

mafredri:
> Support older zsh versions (5.0.x) where `zpty` doesn't give a file descriptor.

Remembered you had this goal here, so I thought I'd inform you I've figured out a way to do this in https://github.com/mafredri/zsh-async/pull/13. So far it has worked reliably for me on a bunch of different systems and versions of zsh.

ericfreese:
Thanks for all the notes here @mafredri! I have been busy with holidays and other projects for a while but I'm hoping to get back to this soon. I've just started implementing high-level integration tests using tmux and RSpec (see [tmux_integration_testing](https://github.com/zsh-users/zsh-autosuggestions/tree/tmux_integration_testing) branch) to hopefully have a more robust test framework in place before continuing this async work. Will definitely dig into your comments more as I get back into this 😄

mafredri:
@ericfreese you're welcome! Just a FYI, `zpty` also works very well for integration testing and allows you to stay completely within shell (e.g. no need for ruby dependency), for an example you can look at the zsh-async [zle test](https://github.com/mafredri/zsh-async/blob/e702ec4697ab6469b8948d599768227304d146b1/async_test.zsh#L434-L493), inspired by zsh [comptest](https://github.com/mafredri/zsh-async/blob/e702ec4697ab6469b8948d599768227304d146b1/async_test.zsh#L434-L493). Not telling you how to write your tests though, so keep on doing it as best fits you 😄 .

src/async.zsh
ericfreese: So I'm playing with `zpty -r` over here, and I'm seeing some non-printable characters showing up in my output. 

screenshot 2017-01-24 17 51 09

screenshot 2017-01-24 17 51 21

You ever seen this? I can probably remove them with a regex or something, but I don't quite understand where these chars are coming from.

ericfreese: Was able to get rid of the carriage return by running `stty -onlcr` in the pty

ericfreese:
@z0rc @balta2ar @mafredri This is getting pretty close. Would you guys be able to pull this branch down and play around with it a bit? See if you can break things? Could use some feedback.

z0rc:
@ericfreese I switched to this branch. So far so good. There is one thing I noticed. I'm using midnight commander and when open its subshell there is `_mc_precmd(){ pwd>&8;kill -STOP $$ }; precmd_functions+=(_mc_precmd)` in prompt as a suggestion. It looks like this:
image
This command is executed by mc itself on start to sync subshell `pwd` with mc. More interestingly, this suggestion is visible only one out of two times. It works like this:
* open `mc`
* `ctrl+o` to open subshell, suggestion is present
* `ctrl+o` to return to mc
* `ctrl+o` to open subshell again, suggestion is gone
* `ctrl+o` and `ctrl+o`, suggestion present
* `ctrl+o` and `ctrl+o`, suggestion is gone
* ...

ivan:
I am testing e3eb286ea213dbcac4f9c22f37ba084dd9e22024 now and my silence will indicate that I didn't have any problems 😄

ivan:
I noticed a problem with e3eb286ea213dbcac4f9c22f37ba084dd9e22024 that I never saw with fedc22e9bbd046867860e772d7d6787f5dae9d4c (and could not reproduce when I reverted to it): the autosuggestion sometimes fails to become invalidated and starts crawling to the right: https://ludios.org/tmp/zsh-autosuggestions-async-issue.mp4

I'm using zsh 5.1.1-1ubuntu2 over ssh on Ubuntu 16.04 with these zsh-autosuggestions options: https://gist.github.com/ivan/79de5e87210e8cf21e305bb4c30c4360#file-zshrc-L161-L167

ericfreese:
Thanks @ivan and @z0rc 👍  I'll look into these issues

mafredri:
I'm seeing stderr output in the suggestions, a `exec 2>/dev/null` inside the worker should help here. 

Another, more worrisome, thing I noticed is that e3eb286 breaks `zsh-async`. I've not investigated it thoroughly, but I believe it's due to the constant deletion and recreation of the asug worker. Unfortunately `zpty` is not a very stable zsh module and is pretty much a hack (from what I gather). Whenever a `zpty` instance is deleted (in this case the asug worker) all things running in any `zpty` receive a `SIGHUP`. I capture this signal in the async worker, so it survives, but all child jobs are killed, effectively nullifying `zsh-async` 😞 .

Technically this means that in the current state, zsh-asug can't co-exist with anything else using `zpty` and might even stand in the way of the user using `zpty`. This is really the fault of the `zpty` module though, and one of the reasons for `zsh-async`s existence is to circumvent these kinds of limitations in `zpty`. I guess it's due time to report this on the `zsh-workers` mailing list.

balta2ar:
I'm on this branch for only a day, so I need more time. Before that I was on the async pull request implemented using zsh_async. From the end user point of view, I see no difference, only that maybe now the shell is slightly faster (it matters when I traverse dirs in MC).

One thing that really annoys me though is that [z](https://github.com/rupa/z) looses its history when I press Enter rapidly multiple times (I sometimes clear my window this way), before zsh is able to prepare another prompt. This is 100% reproducible on my home machine, but not on my work machine. I don't know whether the culprit is z, fzf wrapper of zsh async, I'm just mentioning it here as zsh async may be related (I'm not sure).

z0rc:
Confirming @mafredri issue, using this version of z-asug breaks [pure](https://github.com/intelfx/pure) prompt I'm using, because this prompt heavily relies on zsh-async. I'm switching back to https://github.com/PythonNut/zsh-autosuggestions for time being.

ericfreese:
@mafredri

> I'm seeing stderr output in the suggestions

What is the error output you're seeing?

> Whenever a `zpty` instance is deleted (in this case the asug worker) all things running in any `zpty` receive a `SIGHUP`.

That seems strange. Any idea why it does that? Bug? Feature?

> Technically this means that in the current state, zsh-asug can't co-exist with anything else using `zpty` and might even stand in the way of the user using `zpty`.

I'll look into this. Would definitely be a deal-breaker for using `zpty` for async...

ericfreese:
@balta2ar 

> One thing that really annoys me though is that z looses its history when I press Enter rapidly multiple times

To confirm, this does not happen with @PythonNut's fork? Anything you can do to nail down why this is happening would be very helpful.

mafredri:
@ericfreese 

> What is the error output you're seeing?

The specifics don't really matter, in my particular case it was a failure to set `nice` on a system where this isn't supported (happens when spawning a child process), point being that any unexpected error could occur and it's not a nice user experience to see it :).

> That seems strange. Any idea why it does that? Bug? Feature?

Pretty sure it's a bug. I've looked through the `zpty.c` source but couldn't find the reason for it happening. Maybe some shortcut taken while forking the process.

---

@balta2ar

I've experienced the same issue with `z`. It's definitely a bug/race-condition in `z`. My theory is that one instance of `z` manages to read the database at some point where it's being replaced by another instance, resulting in an empty db.

For this reason I made sure that `zsh-async` does not run any hooks inside the worker ([async.zsh#L75-L80](https://github.com/mafredri/zsh-async/blob/e702ec4697ab6469b8948d599768227304d146b1/async.zsh#L75-L80)).

I believe this issue was aggravated by either https://github.com/rupa/z/commit/4a8b741d1941c7a6a38b2e53292ed4758a2a39bd or https://github.com/rupa/z/commit/1a361f6dc10e54e34e42f1e03850692959c6b413 or both.

ericfreese:
@mafredri 

After a bit of digging...

It looks like `zpty -d` was modified from sending SIGHUP to the single process to sending it to the process group with commit [4852545](https://github.com/zsh-users/zsh/commit/4852545#diff-4e3520b4674e72abe40a93f0f6ece66dR367).

Looks like this is the email that initially added the `deleteptycmd` function: http://www.zsh.org/mla/workers//2000/msg00216.html

... and here's the email that changed it to send the signal to the whole process group: http://www.zsh.org/mla/workers//2000/msg00850.html

Will dig into this more later...

mafredri:
Thanks for digging @ericfreese, logically the SIGHUP should be the culprit but looking at the code I don't see how. While I investigated this before I came to the conclusion that the `zpty` process should indeed have a new process group, and thus the parent, and the other `zpty`'s, should be unaffected. My knowledge of C and unix processes is limited though, so I might be missing something.

I'll try to look into this some more as well.

ericfreese:
@mafredri 

Interestingly, I can't recreate the behavior using the following:

```shell
zsh -f
% zmodload zsh/zpty
% zpty -b cat cat
% zpty -b dog cat
% zpty -d cat
% zpty -t dog; echo $?
0 # <= Would have gotten 1 if dog had also gotten the SIGHUP
```

mafredri:
@ericfreese you stumbled upon the one case where deleting a `zpty` does not affect others, i.e. deleting the first (or an earlier) one. If you did `zpty -d dog` instead, and then `zpty -t cat` you would see that cat was killed by the deletion of dog (since dog came after). The newest `zpty` will send the SIGHUP to all other `zpty`'s that are active, but the oldest will only kill itself :-/.

ericfreese:
It looks like specifically, deleting a zpty will cause SIGHUP to be sent to that zpty process and all other zpty processes started _before_ the targeted zpty.

```shell
% zmodload zsh/zpty
% zpty -b cat cat; zpty -b dog cat; zpty -b rat cat
% zpty
(86048) rat: cat
(86046) dog: cat
(86045) cat: cat
% zpty -d dog
% zpty -t cat; echo $?
1
% zpty -t dog; echo $?
zpty: no such pty command: dog
1
% zpty -t rat; echo $?
0
```

I was stepping through with gdb last night and it looks like it has something to do with the [closing of the file descriptor in `deleteptycommand()`](https://github.com/ericfreese/zsh/blob/master/Src/Modules/zpty.c#L490) rather than the [`kill()` call that sends the SIGHUP](https://github.com/ericfreese/zsh/blob/master/Src/Modules/zpty.c#L494). All prior zpty processes become zombies after stepping over the line that closes the fd for the autosuggest zpty.

I wonder if the fd's aren't getting set up right when the zpty processes are being forked... but I too have a limited knowledge of C and unix processes... :\

ericfreese:
Ok so I've got a fix for upstream [here](https://github.com/ericfreese/zsh/commit/058ea4ee0a979f2da91b5feb2dd6d5eb684ede3d). Will get a patch submitted soon.

Turns out the problem was the zpty exit hook getting called from forked zpty processes, deleting all the pty commands that the forked process knew about (all the ones created before it). Fix is to remove the hook from the forked processes right after the split. We only want the hook to be called with the original zsh process exits.

Next step is to try to figure out a workaround until that patch gets released...

mafredri:
Nice catch, indeed deletion of the hook makes sense as the hook being active explains the behavior we're seeing.

To fix this on any older version of zsh, I can only think of two options:
1. Never deleting a zpty
2. Inactivate the zpty module or it's exit hook inside the zpty clone

For 2. I don't currently have any good leads. Maybe if we somehow could clear the `zpty` module inside the clone it would not know of the others, not sure that's plausible though.

ericfreese:
@mafredri It looks like the patch got merged: https://github.com/zsh-users/zsh/commit/caddeca1ac638137b26735fc8c63d08c83be6a90

In the meantime, I think I've got a workaround. And it should work for `zsh-async` too. The ["zshexit" hook fires before the "exit" hook set by the `zpty` module](https://github.com/zsh-users/zsh/blob/master/Src/builtin.c#L5574-L5575). We can set a "zshexit" hook in the forked pty process so that it SIGKILLs itself _just_ before it makes it to the "exit" hook. It's not pretty, but I think it works...

The only thing that comes between the hooks and the actual `exit()` is some job cleanup. So I unset `MONITOR` which I think would prevent anything from creating a background job.

```diff
diff --git a/src/async.zsh b/src/async.zsh
index 40de015..fd92383 100644
--- a/src/async.zsh
+++ b/src/async.zsh
@@ -7,12 +7,22 @@
 _zsh_autosuggest_async_server() {
 	emulate -R zsh
 
+	# Do not allow job control, since we'll be terminating this
+	# process early and won't be able to properly clean them up
+	unsetopt MONITOR
+
 	# Output only newlines (not carriage return + newline)
 	stty -onlcr
 
 	local strategy=$1
 	local last_pid
 
+	dont_kill_other_ptys_please() {
+		kill -KILL $$
+	}
+
+	add-zsh-hook zshexit dont_kill_other_ptys_please
+
 	while IFS='' read -r -d $'\0' query; do
 		# Kill last bg process
 		kill -KILL $last_pid &>/dev/null
```

mafredri:
@ericfreese that's great! (For the future 😔)

So, I did a quick test with your proposed changes to zsh-asug and so far it seems to prevent zsh-asug from conflicting with zsh-async.

Maybe the kill could be changed to `kill -KILL -$$` so that the entire process group is informed, or perhaps a slightly more graceful version: `kill -HUP -$$; kill -KILL $$`.

Out of curiosity, what problems did you encounter with this approach?

ericfreese:
@mafredri 

> Maybe the kill could be changed to `kill -KILL -$$` so that the entire process group is informed, or perhaps a slightly more graceful version: `kill -HUP -$$; kill -KILL $$`.

I don't _think_ this is necessary. A little testing on my system shows sending `SIGKILL` to the process group leader effectively terminates all the processes in the group. I think I'll leave this out for now, but keep it in the back of my mind if we run into issues where this might help in the future.

> Out of curiosity, what problems did you encounter with this approach?

Honestly... I don't remember now that it's been a couple weeks since I made that comment. Everything seems to work fine now.

I've also addressed silencing stderr output from the zpty server process. Mind taking another look and verifying that it is indeed fixed?

---

@z0rc

I wasn't able to reproduce the `mc` issue you were seeing, but I've pushed up a lot of changes since then. Would you mind trying the newest version to see if the problem persists?

---

@ivan 

The suggestions should work much better now. Please try the newest version and let me know if it solves the problems you were seeing earlier.

---

I think we're getting pretty close here! Would love another round of feedback 🙏

z0rc:
Well, with this branch the git status in [pure prompt](https://github.com/intelfx/pure) is broken. It uses zsh-async for fetching all git information and puts it in placeholder. When I switched to this branch the placeholder is always empty. Sorry, but I think I'll be sticking with z-asug async fork for a while.

I'm on macOS 10.12 and zsh 5.3.1 installed via brew if this matters.

I don't want to be rude, but given amount of code duplication and hacks in place I believe it would be more viable to use zsh-async rather than reimplementing the wheel here.

mafredri:
@z0rc that pure fork has an old version of `zsh-async` tho, have you tried using a more recent version?

z0rc:
@mafredri I'm not using async.zsh from that fork and load it directly from https://github.com/mafredri/zsh-async.

ericfreese:
@z0rc Thank you for the feedback, but I do think we're very close here. I've fixed the issue you reported with the `pure` prompt. I'd ask that you bear with me and try it again, but I understand if you're no longer interested, and I can stop leaning on you for testing if you'd like.

@mafredri In case you're curious, I had to also add the `zshexit` hack to the `zpty` created during feature detection c4bfd8e. The tests were also not catching it because they were not testing zptys created _before_ sourcing the plugin.

ericfreese:
Went ahead and merged this into `develop`. We can continue testing and adding fixes on that branch.

PythonNut:
@ericfreese this seems to break my own usage of `zsh-async`. Any idea what's going on?

ericfreese:
@PythonNut Can you verify the commit sha that you're using? There is a bug in `zpty` that we're trying to work around, but the workaround may not be working right. Would be helpful to have more details on what exactly is breaking for you if possible.

mafredri:
@ericfreese interestingly, I was fine on 06fca77ffb7d85c8b83d5daa591e3f30e646f1d5 but after testing develop (4321fc097c408f6a8dfab8acf36d4706adb35386), it's not behaving well with Pure / zsh-async.

Also ran into this after a while:

```console
_zsh_autosuggest_async_pty_create:zpty:12: pty command name already used: zsh_autosuggest_pty
_zsh_autosuggest_async_pty_create:zle:22: Bad file descriptor number for -F: _zsh_autosuggest_async_response
_zsh_autosuggest_async_pty_create:zpty:12: pty command name already used: zsh_autosuggest_pty
_zsh_autosuggest_async_pty_create:zle:22: Bad file descriptor number for -F: _zsh_autosuggest_async_response
```

mafredri:
Actually, a bisect shows 4321fc097c408f6a8dfab8acf36d4706adb35386 to be the offending commit, which seems weird.

ericfreese:
@mafredi You rock, thanks for the bisect. I was also intermittently seeing the `pty command name already used` errors and pushed a few changes up to `develop` to reduce the amount of stuff we were doing on every precmd. For example, we don't need to run feature detection every precmd, and it was probably a mistake to run `async_start` every precmd too. Hope that helps.

balta2ar:
> I've experienced the same issue with z. It's definitely a bug/race-condition in z

@mafredri Are you still using z? Did you manage to come up with something about that? I ended up making regular backups of my ~/.z and adding functions to manually restore it. Of course, this is silly and it's not solving the root cause, but at least I can quickly restore ~/.z.

mafredri:
@ericfreese I noticed another issue where zsh-asug still can interfere with other `zpty`s, even with the `zshexit` hook.

What I think can happen is that zsh-asug creates and deletes it's `zpty` in rapid succession, causing the HUP to go through to other `zpty`s.

Example:

```diff
diff --git a/zsh-autosuggestions.zsh b/zsh-autosuggestions.zsh
index 34ca080..dc5f7c2 100644
--- a/zsh-autosuggestions.zsh
+++ b/zsh-autosuggestions.zsh
@@ -529,6 +529,7 @@ _zsh_autosuggest_async_server() {
 		(
 			local suggestion
 			_zsh_autosuggest_strategy_$ZSH_AUTOSUGGEST_STRATEGY "$query"
+			sleep 0.5
 			echo -n -E "$suggestion"$'\0'
 		) &
 
@@ -589,6 +590,7 @@ _zsh_autosuggest_async_pty_destroy() {
 }
 
 _zsh_autosuggest_async_pty_recreate() {
+	sleep 0.5
 	_zsh_autosuggest_async_pty_destroy
 	_zsh_autosuggest_async_pty_create
 }
```

Here, the sleep inside `_zsh_autosuggest_async_server` allows me to easily trigger the issue with Pure/zsh-async.

The second sleep in `_zsh_autosuggest_async_pty_recreate` essentially fixes the issue, the other `zpty`s are no longer HUPed.

Now I'm not 100% on the root issue, I'm just seeing the correlation here.

---

@balta2ar I do use `z` but I, as you, am living with it and backing up my `.z` =/. Haven't taken the time to properly investigate the issue.

ericfreese:
Thanks @mafredri. I've added it to the todo list on #218.

ericfreese:
@mafredri I was just trying this out and I'm not able to reproduce the issue.

I'm running `pure` [`291574e`](https://github.com/sindresorhus/pure/commit/291574e652c18f3513524be88913a96038e93e19) and `z-asug` c52c428 with `sleep 0.5` added to `_zsh_autosuggest_async_server`.

Do you have any more information on how you were able to reproduce this? A failing spec would be awesome (relevant spec is [`client_zpty_spec`](https://github.com/zsh-users/zsh-autosuggestions/blob/develop/spec/integrations/client_zpty_spec.rb)).

mafredri:
@ericfreese are you inside a git folder with e.g. changes when testing?

When I have the sleep there, the preprompt is never updated with a `*` to indicate dirty status.

Should be even more easily visible / reproducible with https://github.com/sindresorhus/pure/pull/273 because it relies on async even more.

ericfreese:
> are you inside a git folder with e.g. changes when testing?

Yeah I was manipulating a git repo, staging, committing, resetting, etc. Everything seemed to work as expected (including seeing the `*` to indicate dirty status). The output of `zpty` also showed that all `zpty`s were still alive (not terminated).

> Should be even more easily visible / reproducible with sindresorhus/pure#273 because it relies on async even more.

I'll try with this branch to see if I can reproduce.

mafredri:
@ericfreese sorry, missed your edit. I can reproduce on macOS (`10.12.3 (16D32)`) with zsh `5.3.1` from Homebrew, but not with the current `HEAD`.

mafredri:
@ericfreese [here's a recording](https://asciinema.org/a/4xlwcx07qr85jsx40em5mk1og) demonstrating the issue.

I do have other plugins as well, e.g. `zsh-syntax-highlighting`, `zsh-history-substring-search`, but I can easily reproduce without them.

PR #246, opened 2017-05-20
Add 'emacs-forward-word'

fd0:
This commit adds the 'emacs-forward-word' widget to the list of partial accept widgets.


fd0:
Thanks!

PR #257, opened 2017-07-17
Fix conditionals when [ is rebound to a command

belak:
This was reported here:

https://github.com/sorin-ionescu/prezto/issues/1382

I traced it down to this line:

```
if [ -z "$ZSH_AUTOSUGGEST_BUFFER_MAX_SIZE" -o $#BUFFER -le "$ZSH_AUTOSUGGEST_BUFFER_MAX_SIZE" ]; then
```

What I think is happening is that when $ZSH_AUTOSUGGEST_BUFFER_MAX_SIZE is zero-sized (and the single bracket has been redefined), the argument doesn't get passed in, so the test is essentially this:

```
if [ -z -o $#BUFFER -le ]; then
```

Needless to say, that doesn't work as expected.

I replaced all instances of [ to either [[ or (( and simplified some of the syntax. The test is a super dumb test which essentially checks if there's any extra text when typing a dummy string... and the only thing that changes is that [ is rebound to point to the test command.


belak:
I'm stuck a bit on the failing tests and I'm not sure what would be causing that... help would be appreciated.

spec/integrations/rebound_bracket_spec.rb
```diff
@@ -0,0 +1,14 @@
+describe 'rebinding [' do
+  context 'initialized before sourcing the plugin' do
+    before do
+      session.send_string("function [ { test \"$@\" }")
```
belak: I think I just realized that this is probably the problem - there will be an additional argument (`]`) in most of these cases. I'll get this fixed.

jirutka:
Have you tried the following? 

```sh
if [ -z "$ZSH_AUTOSUGGEST_BUFFER_MAX_SIZE"] || [ $#BUFFER -le "$ZSH_AUTOSUGGEST_BUFFER_MAX_SIZE" ]; then
```

belak:
Turns out the error was really just because `make test` wasn't causing the main file to get compiled in CI (I was avoiding committing it to make cleaner diffs and avoid potential conflicts).

psprint:
Just sharing some data – I've compared speeds of conditionals. Bare `[[` is the fastest, however there are 100k repeats, so actual gains are disputable. However, having to somehow choose between `[` and `[[`, this is an actual argument for `[[`. Redefining `[` is also a problem – in Zplugin I once had to quickly do `[` -> `builtin [` in each `if` etc. to help user. This points to another fact:

```SystemVerilog
% type -w '['
[: builtin
% type -w '[['
[[: reserved
```
So using `[` is like calling `echo`.

The speed comparison:

https://www.zsh.org/mla/users/2016/msg00783.html

spec/integrations/rebound_bracket_spec.rb
```diff
@@ -0,0 +1,14 @@
+describe 'rebinding [' do
+  context 'initialized before sourcing the plugin' do
+    before do
+      session.send_string("function [ { $commands[\\[] \"$@\" }")
+      session.send_keys("'enter'")
```
ericfreese: You should be able to use `session.run_command` here instead of both `send_string` and `send_keys`.

https://github.com/zsh-users/zsh-autosuggestions/blob/master/spec/terminal_session.rb#L25-L30

ericfreese:
Manually rebased, squashed, and merged as commit 940e10a691b34f0281d09794135e8ab81e8fae32.

Thank you very much for your work here @belak!

belak:
Thanks for taking a look! It'll be nice to update prezto to include this once there's a release with it. :)

ericfreese:
> It'll be nice to update prezto to include this once there's a release with it.

Yes, I think it's about time to cut a 0.4.1 release. I'll try to get that together in the next few days.

ericfreese:
@belak v0.4.1 is now tagged and merged into master.

belak:
Awesome, thanks for the quick work! It's up to date in prezto now.

Issue #6, opened 2013-11-05
Autosuggestions do not work with zsh-syntax-highlighting


ZyX-I:
It is somewhat expected since both plugins are using the same zle widgets. I guess you need to cooperate and work out some standard that will allow using both plugins.

Fish has both completion and highlighting and it looks nice. I do not like much more limited capabilities of it though.

balta2ar:
@tarruda I'm experiencing the same problem :( Have to live on detached HEAD because up-to-date versions of zsh-autosuggestions and zsh-syntax-highlighting don't play well together. Is there any chance you could look into this problem once again, please?

tarruda:
> Is there any chance you could look into this problem once again, please?

Not in the foreseeable future, sorry.

faceleg:
@balta2ar we need someone with time & zsh experience to refactor some of this :8ball:

santiycr:
Just in case others are struggling with this one while using prezto. If when you submit commands with active auto-suggestions, they loose their highlighting and get confusing with the actual command that was run:
![2016-05-19_16-53-36](https://cloud.githubusercontent.com/assets/105375/15435674/8e27ec12-1e72-11e6-8ff8-77f1db5c2a0d.png)

(notice, I had typed and submitted `git a`, while the rest was grayed out before submitting)

The following `syntax-highlighting` cleared the issue for me:
https://github.com/sorin-ionescu/prezto/tree/master/modules/syntax-highlighting#highlighting

iamsandesh:
I was struggling with this issue, then I loaded the 'autosuggestions' module after the 'syntax-highlighting' module then both the modules seems to be working fine.

I also enabled the coloring for 'autosuggestions' by uncommenting the below line
 zstyle ':prezto:module:autosuggestions:color' found 'blue'

schlichtanders:
@iamsandesh that sounds really good! can you explain where to find this line?

iamsandesh:
I am updating it in the "~/.zprezto/runcoms/.zpreztorc" file which is present in $HOME folder. 

The line is already present in the file but commented out, so the line needs to be uncommented and the color has to be specified in that line as it is not already specified.

psprint:
I've debugged my `fast-syntax-highlighting` (`z-sy-h` fork) and it's like it is described in #260 – `builtin zle "$@"` returns `1` and `_zsh_highlight` isn't run. So I patched my `f-sy-h`:

```diff
 _zsh_highlight_call_widget()
[...]

PR #260, opened 2017-07-23
Compatibility with zsh-syntax-highlighting

toadjaune:
Hello everybody,

This pull request tries to fix https://github.com/zsh-users/zsh-autosuggestions/issues/6.

As described there, we currently have compatibility issues between zsh-autosuggestions and zsh-syntax-highlighting.

Issue description & reproduction
--------------------------------

Start a zsh session with both zsh-autosuggestions and zsh-syntax-highlighting sourced.

Currently, the symptoms are the following :

- Whenever we accept the full suggestion (for example with `vi-forward-char`, ie the right arrow in vi mode), the line gets completed, but the highlighting is not updated.
- Whenever we accept & execute the full suggestion, the highlighting gets correctly updated (I tested this with a manual binding to the `autosuggest-execute` widget)
- Whenever we accept a partial suggestion, the highlighting gets correctly updated (Tested this with `vi-forward-word` only)

Explanation
-----------

So, I tracked the issue down the code of both plugins, here is what I understood so far :
- No matter what, zsh-autosuggestions is going to wrap the widgets _after_ zsh-syntax-highlighting, because zsh-syntax-highlighting does it when it is loaded, while zsh-autosuggestions does it at runtime with a `precmd` hook (I wonder why).
This is what we want, because we need zsh-syntax-highlighting to do its job between the initial reset of highlighting, and the highlighting of the suggestion.
- Now, here is what happens when we autocomplete. I'm taking the example of `vi-forward-char`, but something similar probably happens with other widgets accepting the full suggestion :
  - [`_zsh_autosuggest_accept`](https://github.com/zsh-users/zsh-autosuggestions/blob/master/src/widgets.zsh#L137) gets triggered
  - Among other things, the cursor is moved at the end of the line
  - The original widget (In this case, the version of `vi-forward-char` wrapped by zsh-syntax-highlighting)
  - The execution reaches [`_zsh_highlight_call_widget`](https://github.com/zsh-users/zsh-syntax-highlighting/blob/master/zsh-syntax-highlighting.zsh#L266), which himself calls `builtin zle .vi-forward-char`
  - `vi-forward-char` fails. Although the return code of `vi-forward-char` does not seem to be [documented](http://zsh.sourceforge.net/Doc/Release/Zsh-Line-Editor.html#Movement), my guess is that since it cannot move to the next character, it considers this invocation to be a failure (Which would make sense).
  - `_zsh_highlight` never gets called.

Solutions
---------

This PR implements a solution, it manually calls `_zsh_highlight` (if defined) in `_zsh_autosuggest_accept`.
While this solution seems to work perfectly (so far), it seems really hacky to me.
I don't really like the idea of having one plugin referencing an internal function of the other.

So, after thinking about it, I had another idea, but I wanted some advice before trying to implement it :
Instead of manually moving the cursor in `_zsh_autosuggest_accept`, we could do it through another widget, such as `vi-end-of-line` (with a count in case of multiple lines).
Since this widget is wrapped by zsh-syntax-highlighting, and since it would _actually do_ something, this approach should transparently solve our issues.
Even better, this is a generic approach that may solve issues of interaction with other plugins wrapping the widgets.

So, what do you guys think ?


z0rc:
I'm not a maintainer, here is general recommendation.

Instead of spawning subshell with quotes and testing its return code in `if` clause, it's better to check the exit code of `whence -w` directly in `if`.

Issue #258, opened 2017-07-18
_zsh_autosuggest_strategy_match_prev_cmd:17: bad math expression: operator expected at ...


mconigliaro:
To reproduce, start typing: `echo $(basename $PWD)`

```
$ echo $(basename $P
_zsh_autosuggest_strategy_match_prev_cmd:17: bad math expression: operator expected at `$(basename...'
$ echo $(basename $PP
```

isaacwhanson:
this is the same problem as with #247 .  I have a *possible* solution there.

ericfreese:
Should be fixed on `develop`.

Issue #247, opened 2017-05-21
Breakage after typing a [


ivan:
This happens with 0.4.0, but not 0.3.3:

```
at@ra:/home/at# while [       
at@ra:/home/at# cd bin                    
at@ra:/home/at/bin:master# while [       
_zsh_autosuggest_strategy_match_prev_cmd:17: bad math expression: operator expected at `bin'
at@ra:/home/at/bin:master# while [
at@ra:/home/at/bin:master# while [       
_zsh_autosuggest_strategy_match_prev_cmd:17: bad math expression: operator expected at `bin'
at@ra:/home/at/bin:master# while [
at@ra:/home/at/bin:master# [
_zsh_autosuggest_strategy_match_prev_cmd:17: bad math expression: operator expected at `bin'
at@ra:/home/at/bin:master# [
at@ra:/home/at/bin:master# [
_zsh_autosuggest_strategy_match_prev_cmd:17: bad math expression: operator expected at `bin'
at@ra:/home/at/bin:master# [
at@ra:/home/at/bin:master# [
_zsh_autosuggest_strategy_match_prev_cmd:17: bad math expression: operator expected at `bin'
at@ra:/home/at/bin:master# [
at@ra:/home/at/bin:master# ]]]][
_zsh_autosuggest_strategy_match_prev_cmd:17: bad math expression: operator expected at `bin'
```

The `_zsh_autosuggest_strategy_match_prev_cmd` error pops up immediately after pressing the `[` key.

```
# zsh --version                                               
zsh 5.1.1 (x86_64-ubuntu-linux-gnu)
```

.zshrc settings: https://gist.github.com/ivan/79de5e87210e8cf21e305bb4c30c4360#file-zshrc-L200-L206 possibly-relevant bit being

```
ZSH_AUTOSUGGEST_STRATEGY=match_prev_cmd
```

ericfreese:
Looks like this is specific to `match_prev_cmd` strategy.

Birne94:
I am receiving a similar error with the latest release (master @ 2cb6eb6 - (tag: v0.4.0)) when pressing the number sign (#):

```
daniel@Daniels-MBP ~/Workspace/dotfiles % #
_zsh_autosuggest_strategy_match_prev_cmd:17: bad math expression: operator expected at `ZSH_AUTOSU...'
daniel@Daniels-MBP ~/Workspace/dotfiles % #
```

paulbdavis:
[...]

PR #266, opened 2017-08-18
escape prefix before matching in match_prev_cmd strategy

isaacwhanson:
escape shell characters in the `prefix` variable before matching history.  Should solve issues #247 and #258.


ericfreese:
Thanks @isaacwhanson! I've added a test for this issue on the [`fixes/match_prev_cmd_special_chars`](https://github.com/zsh-users/zsh-autosuggestions/commit/9e110406fa743b07ecc539c9b4ecab0e9bb50449) branch and it looks like your patch doesn't quite fix all the cases. Would you mind taking another look and submitting a new PR to that branch that makes the test pass?

isaacwhanson:
I will give it a shot :)

PR #267, opened 2017-08-18
escape pattern-matching chars on $prefix for match_prev_cmd strategy

isaacwhanson:
this should fix issues #247 and #258 where pattern-matching characters in the buffer are not escaped before matching against history when using the match_prev_cmd strategy.


isaacwhanson:
note I had to use `${(b)prefix}`, not `${(q)prefix}` as previously submitted.

ericfreese:
Thanks @isaacwhanson! :+1:

Issue #248, opened 2017-05-21
Autosuggestion is not invalidated when typing a * or ?


ivan:
I'm having this problem with 0.4.0, but not 0.3.3: when the typed command includes a `*` or `?`, the autosuggestion is not invalidated and starts crawling to the right: https://ludios.org/tmp/zsh-autosuggestions-async-issue.mp4

```
# zsh --version                                               
zsh 5.1.1 (x86_64-ubuntu-linux-gnu)
```

.zshrc settings: https://gist.github.com/ivan/79de5e87210e8cf21e305bb4c30c4360#file-zshrc-L200-L206

(Originally reported at https://github.com/zsh-users/zsh-autosuggestions/pull/180#issuecomment-275911411.)

ericfreese:
Only occurs with `match_prev_cmd` strategy.

ericfreese:
I believe this is fixed on `develop`. Please try it out @ivan

PR #270, opened 2017-09-10
Simplify escaping of pattern and fix match_prev_cmd strategy

ssiegel:
Maybe this is also a fix for #247, #248 and #258. Supersedes #267.

Testcase:
Using match_prev_cmd strategy and with these lines in history:
echo '1^'
echo '2^'
echo '1^'

type:
echo       (unexpected suggestion echo '1^' instead of echo '2^')
echo '1^1  (wrong suggestion echo '1^1echo '1^')
echo '1^#  (error "bad math expression")


ericfreese:
Looks great, thanks!

PR #305, opened 2018-01-15
Behave correctly in environments without zsh/zpty

okdana:
The plug-in currently assumes that `zsh/zpty` has been successfully loaded (and the `zpty` built-in made available), even when the asynchronous-fetch feature isn't enabled. The problem with this is that the module is completely optional, and some systems (particularly those with a statically linked zsh) don't have it at all. In those environments, the plug-in is almost unusable, because it's constantly dumping garbage to the screen about `zpty` being missing.

This could probably be handled more elegantly in general, but the change i've made here is super easy and at least makes it usable for me.

(Note: `zpty` now has an `-s` option to suppress errors without messing with redirections, but i think it's only available in 5.3+ — my static zsh is still 5.2.)

### DIFF
--- a/src/setup.zsh
+++ b/src/setup.zsh
@@ -7,4 +7,4 @@
 autoload -Uz add-zsh-hook
 
 # Asynchronous suggestions are generated in a pty
-zmodload zsh/zpty
+zmodload zsh/zpty 2> /dev/null || zpty() { return 1 }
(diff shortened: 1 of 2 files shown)


okdana:
Changed the original source on this one too. Rebased

ericfreese:
I'm actually going to close this as I've implemented async suggestions without zpty on `develop`. Going forward, zpty is only required for `completion` suggestion strategy, and it will fail gracefully (not offer a suggestion) if zpty is not available. See PR #348.

Thanks for bringing this to my attention though and for providing a PR :)

PR #315, opened 2018-03-05
Add new "predefined" strategy for zsh-autosuggestions

skywind3000:
After using and updating it for nearly a month, I believe the new strategy is stable now and could be trusted.

So, this PR has been created. and wish other zsh-autosuggestions users could benefit from it.

### DIFF
--- a/src/strategies/predefined.zsh
+++ b/src/strategies/predefined.zsh
@@ -38,7 +38,7 @@ _zsh_autosuggest_predefined_generate() {
     for p ("${(@s/:/)PATH}"); do
         [ ! -d "$p" ] && continue
         cd "$p"
-        local files=("${(@f)$(ls -la | awk -F ' ' '{print $9}')}")
+        local files=($(command ls))
         for fn in ${files}; do
             if [ -x "$fn" ] && [[ "${fn:l}" != *.dll ]]; then
                 if [ -f "$fn" ] && [[ "${fn:l}" != *.nls ]]; then
--- a/src/strategies/predefined.zsh
+++ b/src/strategies/predefined.zsh
@@ -65,8 +65,11 @@ _zsh_autosuggest_strategy_predefined() {
     emulate -L zsh
     setopt EXTENDED_GLOB
     local prefix="${1//(#m)[\\*?[\]<>()|^~#]/\\$MATCH}"
+
+	# search the history at first
     local result="${history[(r)${prefix}*]}"
 
+	# search the predefine files if nothing found in history
     if [[ -z "$result" ]]; then
         if (( ! ${+_ZSH_AUTOSUGGEST_PREDEFINE} )); then
             typeset -g _ZSH_AUTOSUGGEST_PREDEFINE=()
--- a/src/strategies/predefined.zsh
+++ b/src/strategies/predefined.zsh
@@ -27,7 +27,7 @@ _zsh_autosuggest_predefined_generate() {
     # skip when ~/.zsh_autosuggest exists
     [ -f "$pname" ] && return
 
-    echo "autosuggestions is generating: $pname"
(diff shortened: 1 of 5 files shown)


src/strategies/predefined.zsh
```diff
@@ -0,0 +1,99 @@
+# zsh-augosuggestion: predefined strategy
+# 1. search in history, returns result if it exists
+# 2. search in predefined files
+#
+# configuration:
+# 
+# ZSH_AUTOSUGGEST_PREDEFINE_NAME  - auto generated predefine file name
+# ZSH_AUTOSUGGEST_PREDEFINE_PATH  - user defined files separated by semicolon
+#
+# "$HOME/.zsh_autosuggest" will be generated at the first time
```
z0rc: Please don't default to dotfile in top level of home directory. [XDG Base Dir spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) is more than a decade old and equally applies to GUI applications and CLI utilities.

`$XDG_CACHE_HOME` is a perfect location for this kind of file.

src/strategies/predefined.zsh
```diff
@@ -0,0 +1,99 @@
...
+#
+# zsh-autosuggestions/predefined.txt is generated from tldr pages and
+# will ship with predefined.zsh .
+#
+
+_zsh_autosuggest_script_path="${0:A:h}"
+
+_zsh_autosuggest_predefined_generate() {
```
z0rc: Seems that this function will lock prompt until it's finished, which is poor user experience. z-asug already has async support, and this task should be started in background.

src/strategies/predefined.zsh
skywind3000: `$HOME/.zsh_autosuggest` has been changed to `$XDG_DATA_HOME/zsh_autosuggest/predefined`, or `~/.local/share/zsh_autosuggest/predefined` if `$XDG_DATA_HOME` is empty.

skywind3000:
`$HOME/.zsh_autosuggest` has been changed to `$XDG_DATA_HOME/zsh_autosuggest/predefined`, or `~/.local/share/zsh_autosuggest/predefined` if `$XDG_DATA_HOME` is empty.

and one last thing: async support

I have investigate `async.zsh` in zsh-asug, it is not the real async, despite that strategies are running in the background, the foreground process will wait the response from background process.

I have tried sleep 5 seconds in the strategy function, it was supposed to be async. but my prompt had been locked for 5 seconds due to the response waiting.

But at the main time, I optimize the performance of `_zsh_autosuggest_predefined_generate` function, and it will take only 150ms to generate the predefined, and I think it is accaptable because it will run only once.

z0rc:
Okay, fair point. But at this case, please make `_zsh_autosuggest_predefined_generate` run only when strategy is explicitly enabled. Because right now seems it's running always, despite the chosen strategy.

z0rc:
Please provide documentation, especially [this](https://github.com/zsh-users/zsh-autosuggestions#suggestion-strategy) part should be updated.

And tests, I believe at current state it's mandatory for this project.

PS I'm not a maintainer, so @ericfreese has the last word on this.

skywind3000:
Done:
- database generating will happen only if `predefined` strategy has been chosen.
- update documentation about `predefined` strategy
- allow use set ZSH_AUTOSUGGEST_STRATEGY before plugin initializing.

src/strategies/predefined.zsh
partounian: @z0rc this plugin does not have correctly working async func but yes locking is poor experience

ericfreese:
Thanks for your PR for this, but I have implemented suggestions based on the completion engine, which I believe will cover this use case without needing to pre-generate anything.

Issue #241, opened 2017-05-16
ZSH_AUTOSUGGEST_USE_ASYNC breaks shell history


austinarbor:
When adding the option `ZSH_AUTOSUGGEST_USE_ASYNC=true` to my `~/.zshrc` file, this breaks command history.  Pressing up button on keyboard will retrieve at most the 2 most recent commands.  When I remove this option from my `~/.zshrc` file, command history functionality returns to normal

ericfreese:
Oh dang. I bet this is related to #227. One of my last commits before `v0.4.0` was to disable async by default so I bet that's what "fixed" it.

cig0:
Hi all,

Can anybody confirm if shell history is broken again?

**Zsh**
```
~ % zsh --version
zsh --version
zsh 5.4.2 (x86_64-apple-darwin17.0.0)
```
**zsh-autosuggestions** (just cloned), enabled via _oh-my-zsh_:
```
~ % head -n3 /Users/cig0/.oh-my-zsh/custom/plugins/zsh-autosuggestions/CHANGELOG.md
# Changelog

## v0.4.0
```
**~/.zshrc**
```
plugins=(
    ...
    zsh-autosuggestions
    ...
)

# Plugins configuration
...
ZSH_AUTOSUGGEST_USE_ASYNC="true"  ## zsh-autosuggest
```

Thanks!

partounian:
Yes it is still broken.

Also, @cig0 you need to set ZSH_AUTOSUGGEST_USE_ASYNC before the plugins array.

partounian:
Also we should probably check if `ZSH_AUTOSUGGEST_USE_ASYNC` is set to `1` or `true`, rather than checking if defined.

aarbor989:
@partounian does it need to be before the plugins array?  I don't see that mentioned in the docs

partounian:
@aarbor989 I was wrong, it executes before each prompt (precmd).


https://github.com/zsh-users/zsh-autosuggestions/blob/master/zsh-autosuggestions.zsh#L699

BirkhoffLee:
This is still broken, b4 it gets fixed i suggest writing this issue to README.md to notice ppl

ericfreese:
I've got a hypothesis based on #227 that this is only an issue when using `up-line-or-beginning-search` and `down-line-or-beginning-search`. Is anyone here seeing this issue while using other widgets?

You can check what your up and down keys are bound to by running:

```
[...]

Issue #227, opened 2017-03-28
zsh-autosuggestions breaks `up-line-or-beginning-search` and `down-line-or-beginning-search`


OmeGak:
I use `up-line-or-beginning-search` and `down-line-or-beginning-search` to navigate history up from the position of the cursor. It used to work just fine but after I upgraded some time ago it stopped working. When I press `<UP>`, I can only go a couple of entries back in the history. When I press `<DOWN>` afterwards, I only get a beep.

I found this configuration, in combination with `match_prev_cmd` strategy, very powerful, as it gave me a way to preview the most recent history result from where I had typed, press `<UP>` and then keep going back in history. If I changed my mind, pressing `<DOWN>` would leave me exactly where I was at the beginning, with the suggestion still there.

I have tested in the `develop` branch and the issue remains.

kylealwyn:
same issue.

Xazzzi:
Same here on `develop`. Any fix available?

OmeGak:
I just updated and it seems to be working fine again.

Xazzzi:
Confirmed, works like a charm! =)

cig0:
Hi all,

Can anybody confirm if shell history is broken again?

**Zsh**
```
~ % zsh --version
zsh --version
zsh 5.4.2 (x86_64-apple-darwin17.0.0)
```
**zsh-autosuggestions** (just cloned), enabled via _oh-my-zsh_:
```
~ % head -n3 /Users/cig0/.oh-my-zsh/custom/plugins/zsh-autosuggestions/CHANGELOG.md
# Changelog

## v0.4.0
```
**~/.zshrc**
```
plugins=(
    ...
    zsh-autosuggestions
    ...
)

# Plugins configuration
...
ZSH_AUTOSUGGEST_USE_ASYNC="true"  ## zsh-autosuggest
```

Thanks!

PR #318, opened 2018-03-15
Remove Async section as it's broken

partounian:
Not sure why we didn't do this earlier oops

https://github.com/zsh-users/zsh-autosuggestions/issues/241

@BirkhoffLee

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,10 +129,6 @@ Widgets that modify the buffer and are not found in any of these arrays will fet
 Set `ZSH_AUTOSUGGEST_BUFFER_MAX_SIZE` to an integer value to disable autosuggestion for large buffers. The default is unset, which means that autosuggestion will be tried for any buffer size. Recommended value is 20.
 This can be useful when pasting large amount of text in the terminal, to avoid triggering autosuggestion for too long strings.
 
-### Enable Asynchronous Mode
-
-As of `v0.4.0`, suggestions can be fetched asynchronously using the `zsh/zpty` module. To enable this behavior, set the `ZSH_AUTOSUGGEST_USE_ASYNC` variable (it can be set to anything).
-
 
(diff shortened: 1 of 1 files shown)


ericfreese:
I think I'd rather just fix async than remove documentation for it.

partounian:
But it has been broken for many months now without any work it seems. I understand your point, and we can all get busy in life for OSS so if something doesn't work, it's probably better to just not advertise it.

ericfreese:
Sorry for the confusion. My comment above probably would have been better worded as:

> I think I'd rather fix this bug with async than remove documentation for the entire feature.

I'm not convinced it's "broken". It looks like there's a bug with it, but the feature is still very usable. Playing around with it a bit, it looks like the default `up-line-or-history` and `down-line-or-history` widgets work fine with async enabled, so it might just be an issue with `up-line-or-beginning-search` as [the first bug report](https://github.com/zsh-users/zsh-autosuggestions/issues/227) states.

I appreciate the effort, but lets focus our energy on fixing that bug instead of rolling back support for the whole feature :) You could start by sharing the details of your .zshrc that cause this issue on that issue thread.

partounian:
https://github.com/zsh-users/zsh-autosuggestions/issues/241

partounian:
I believe you can find more details here https://github.com/zsh-users/zsh-autosuggestions/issues/241 I will give it another test now

ericfreese:
> Though do you have thoughts on checking if the ZSH_AUTOSUGGEST_USE_ASYNC is 1 or true, instead of just set?

I don't have a strong opinion but lean toward just checking if it's set. If we did decide to change it, we should mark the old style as deprecated and display a warning for a while encouraging users to change to `..._USE_ASYNC=1` or whatever. That's not high on my priority list though.

PR #319, opened 2018-03-23
Don't fetch suggestions after [up,down]-line-or-beginning-search

### LINKED ISSUE #241: ZSH_AUTOSUGGEST_USE_ASYNC breaks shell history

ericfreese:
Should fix #241 and #227.

These widgets rely on `$LASTWIDGET` being set to restore the cursor position. When asynchronous suggestions are enabled, and the widget triggers a suggestion to be fetched, `autosuggest-suggest` will be called and $LASTWIDGET will be set to it.


partounian:
I can confirm this is working

Issue #188, opened 2016-08-19
Partial accept not working on vi-forward-word when in vi-mode normal mode


kolyuchiy:
It only accepts until the end of current word but not further.
But if I do `bindkey '^[f' vi-forward-word` then it works correctly in insert mode.

swalladge:
I'm having the same problem. It also seems to buggily work sometimes - for example if I alternate between hitting the keys for vi-forward-blank-word and vi-forward-blank-word-end it will partial accept up to a certain point. It seems to be a problem with being in `vicmd` mode - even doing something like `bindkey -M vicmd '^[f' vi-forward-word` won't work.

IngoHeimbach:
For whatever reason, the cursor position must be increased by one if a vi widget partially accepted a suggestion (see my [patch](https://github.com/IngoHeimbach/zsh-autosuggestions/commit/af25faed003de1557b9e2e74173f4c58ed80101a)). I don't know if this is the proper way of solving this issue but it seems to work.

toadjaune:
Just ran into this bug myself, and I think I found a [cleaner fix](https://github.com/zsh-users/zsh-autosuggestions/pull/324).

The issue is that in vicmd keymap, the cursor is _on_ the last character of the line, instead of right _after_ it. This is an issue since it breaks an assumption made by the code to check whether we went further than the initial buffer size or not.

My fix simply uses an intermediate variable for the check, that is incremented by 1 in `vicmd` keymap, thus restoring the usual behaviour of the test.
Moreover, we don't change the actual CURSOR value, which prevents any weird side-effect from happening.

For some reason that I don't undersand yet, this seems to also fix the issue in `visual` keymap.

ericfreese:
See my comment on @toadjaune's PR here: https://github.com/zsh-users/zsh-autosuggestions/pull/324#issuecomment-388974152

What are people expecting to happen when you use `vi-forward-word` vs. `vi-forward-word-end`?

`vi-forward-word` moves the cursor to the first character of the next word. Would you actually expect that character to be accepted (added to the buffer)?

ericfreese:
Should be fixed on `develop` branch. Will go out with next release.

PR #324, opened 2018-05-04
Fix vi partial accept

toadjaune:
Cf https://github.com/zsh-users/zsh-autosuggestions/issues/188

### DIFF
--- a/spec/integrations/vicmd_partial_accept_spec.rb
+++ b/spec/integrations/vicmd_partial_accept_spec.rb
@@ -2,12 +2,6 @@ describe 'a vicmd mode partial-accept widget' do
   let(:widget) { 'vi-forward-word-end' }
 
   context 'in vicmd mode' do
-#    -> do
-#      let(:before_sourcing) do
-#        session.
-#      end
-#    end
-
     it 'moves the cursor through suggestion as expected' do
       session.run_command("bindkey s vi-cmd-mode")
       with_history('foobar foo') do
--- a/spec/integrations/vicmd_partial_accept_spec.rb
+++ b/spec/integrations/vicmd_partial_accept_spec.rb
@@ -1,0 +1,19 @@
+describe 'a vicmd mode partial-accept widget' do
+  let(:widget) { 'vi-forward-word-end' }
+
+  context 'in vicmd mode' do
+#    -> do
+#      let(:before_sourcing) do
+#        session.
+#      end
+#    end
+
+    it 'moves the cursor through suggestion as expected' do
+      session.run_command("bindkey s vi-cmd-mode")
+      with_history('foobar foo') do
+        session.send_string('fo').send_keys('s').send_keys('e')
+        wait_for { session.content }.to eq('foobar')
+      end
+    end
+  end
+end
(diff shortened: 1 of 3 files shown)


z0rc:
Please use double square brackets (`[[ ]]`) for code consistency.

ericfreese:
Thanks for fixing this! Can you add a test?

toadjaune:
Code style fixed, I'll look into making a test.

toadjaune:
@ericfreese, I can't get the already existing tests to pass : 
`
rspec ./spec/integrations/wrapped_widget_spec.rb:15 # a wrapped widget initialized before sourcing the plugin executes the custom behavior and the built-in behavior
rspec ./spec/integrations/wrapped_widget_spec.rb:32 # a wrapped widget initialized after sourcing the plugin executes the custom behavior and the built-in behavior
`
This happens both on this branch, on develop, and on master.

Is this a known issue ? Is something wrong in my test environment ? Should I just ignore these errors and go ahead adding a test ?

toadjaune:
Here is a test, I don't know if it seems clean enough, though.

ericfreese:
Thanks @toadjaune! I will take a look at this soon.

ericfreese:
I've added [an additional test](https://github.com/zsh-users/zsh-autosuggestions/blob/5ac9363f54b658f234950434f03c540092836f7a/spec/integrations/vi_mode_spec.rb#L23-L36) to the new `fixes/vi-partial-accept` branch that tests a behavior that seems intuitive to me: not accepting the first character of the next word when using `vi-forward-word` to partially accept a suggestion.

With the proposed patch, the new test fails with:

```
  expected: "foobar baz"
       got: "foobar fbaz"
```

What do you think about this case? What do you think would be the most intuitive behavior?

I'm not sure the most elegant way to include the char under the cursor for `vi-forward-word-end`, but not include it for `vi-forward-word`. Ideally both of these tests will pass.

ericfreese:
The more I think about this, the more I think it would be ok for `vi-forward-word` to accept the first char of the next word. You could still accept to just before it by using `f<space>` (which makes me realize I need to add `vi-find-next-char` and `vi-find-next-char-skip` to `ZSH_AUTOSUGGEST_PARTIAL_ACCEPT_WIDGETS`

IngoHeimbach:
In my opinion, the vi widgets should move the cursor like they usually do, so it should be ok to accept the first character of a new word if necessary. In that case, the user can still press `C` do delete this one character and change to insert mode.

ericfreese:
Ok I've applied a similar patch with https://github.com/zsh-users/zsh-autosuggestions/commit/393f7b8bb9d83aef1847d39ad20c6a73849917b8 that should fix issue #188. It's been merged to `develop` branch and will be included in the next release.

Thanks @IngoHeimbach and @toadjaune for your effort on this!

IngoHeimbach:
Thanks everyone for your work! :+1:

toadjaune:
I think that we should probably simply let widgets such as `vi-forward-word` do their job normally, and don't try to change their behavior in any way.

It would be a surprising (and pretty terrible) experience to have the cursor behaving differently depending on whether we're moving through the beginning of the command or through the suggestion. Same goes about behaving differently with and without zsh-autosuggestions.

Therefore, since in `vicmd` keymap, the cursor can never be after the last character, and since `vi-forward-word` will usually bring it on the first letter of a word, I agree that the sanest thing to do here would be … nothing specific (ie let it accept the first letter)

toadjaune:
Oh, I had not reloaded my tab, missed the two latest messages, sorry ^^'

Your solution is probably more elegant than mine, I tried to make something similar but couldn't get it to work.

However, I'm a bit worried that we might encounter issues down the road, by not using $LBUFFER and $RBUFFER, in some weird edge case.

Anyway, thanks everyone, that's a good needle out of our foot :)

Issue #111, opened 2016-02-10
Suggest Program Options/Flags


noctuid:
Fish will suggest program options even if you've never used them before. I'm not sure if this would be easily possible, but it would be really cool. I don't know if there's been an issue for this already, but I couldn't find one when searching.

This may not be as useful for someone using something like auto-fu, but I think it would still be nice.

PythonNut:
This might be doable with [`zsh-capture-completion`](https://github.com/Valodim/zsh-capture-completion).

Hunsu:
> This might be doable with zsh-capture-completion. 

Can you add more details please as I don't know how to use it?

Is there any way to customize zsh-autosuggestions for certain commands? For example don't suggest: `git checkout -b branch`if `branch`already exist.

PythonNut:
@Hunsu ah sorry. Let me clarify: this may be doable in `zsh-autosuggestions` by writing a lot of new code. It's not possible currently. 

Right now, I use [`auto-fu`](https://github.com/hchbaw/auto-fu.zsh) together with `zsh-autosuggestions`. They actually complement each other rather well, although `auto-fu` is quite painful to setup.

ericfreese:
Please try out [`features/completion-suggestions`](https://github.com/zsh-users/zsh-autosuggestions/tree/features/completion-suggestions) branch. See the PR #330 for more info.

noctuid:
I just tried it. It seems to work quite well for some commands (e.g. `git`). On the other hand, I get no suggestions for others like `borg` even though tab completion gives me results.

ericfreese:
I just installed `borg` (had never heard of it before) and I don't get any tab completion results. I'm trying `borg <tab>` and I just get a list of files. I would have expected sub-commands like `init` and `create`.

Do you have custom completion scripts for it?

noctuid:
I do not, but it's possible that my distro (Arch) ships borg with [these](https://github.com/mrkmg/borgbackup-zsh-completion). I'd have to check.

noctuid:
Also, if I type `hg --co` it gives a correct suggestion, but it's preceded by junk characters and then my current input again (e.g. `hg --co^M^M^[[0m^[[23m^[[24m^[[Jhg --co--config`). This happens for anything with `hg`.
`

noctuid:
I've also had my computer completely freeze twice while typing `man ...`. I can't replicate this consistently, and it may be an unrelated issue, but I haven't had this problem before today.

[...]

PR #330, opened 2018-05-16
Features/completion suggestions

ericfreese:
- Adds new `completion` strategy (#111).
- Renames `default` strategy to `history`.
- Makes `ZSH_AUTOSUGGEST_STRATEGY` an array. Strategies will be tried successively and first suggestion found will be used.

`completion` strategy can be pretty slow, so consider turning on `ZSH_AUTOSUGGEST_USE_ASYNC`.

### To Do

- [x] Write a test for it
- [x] Figure out why [CI](https://circleci.com/gh/zsh-users/zsh-autosuggestions/tree/features%2Fcompletion-suggestions) is failing
- [x] Is it possible to suggest non-exported variables? They're not available in the pty.
- [x] Write test for and fix issue with carriage returns in async mode (ex: `echo '^M'` in buffer)
- [x] Fix on older versions of zsh
- [x] Support `ZSH_AUTOSUGGEST_STRATEGY` as an array of strategies to try successively.
- [x] Fix blowing up when single quote entered in buffer.
- [ ] ~Look into (and fix?) leaking pty processes on Mac OSX~ not sure about this one... will wait for confirmed bug reports since it's experimental anyhow.


megalithic:
would LOVE this to be merged.

alok:
I seem to have an issue with this. If I type something like `e` (alias for `$EDITOR`), then suddenly I stop getting completions. It seems to happen if I type in a short command and look for a completion, and lasts for the entire terminal session.

ericfreese:
@alok Interesting, I get more suggestions after aliasing `e` (it suggests `egrep`).

Can you reproduce using `zsh -f` and sourcing the plugin manually? If not, can you find the offending settings in your `.zshrc`?

alok:
Appears to have magically fixed itself.

ericfreese:
Yeah I've run into that. It's on the todo list as "Fix blowing up when single quote entered in buffer."

alok:
FWIW I get leaking processes on the latest release of macOS High Sierra. Not sure how to replicate though.

ericfreese:
Dang I was hoping it was something unrelated on my mac. I'll make an issue to track it. I seem to not have the problem on my linux machine for some reason.

Issue #341, opened 2018-06-18
Problems with setopt warn_create_global


vaeth:
With `setopt warn_create_global` (on branch features/completion-suggestions), I receive for the first usage the warnings
```
_zsh_autosuggest_fetch_suggestion:7: scalar parameter strategy created globally in function _zsh_autosuggest_fetch_suggestion
_zsh_autosuggest_capture_completion:6: scalar parameter REPLY created globally in function _zsh_autosuggest_capture_completion
```
I suppose the former is indeed a missing `local strategy`.
Not sure about the latter; perhaps at least the warning can be suppressed somehow?

ericfreese:
Should be fixed now with 43a011026ffd9c7e9dba4cf022922a4877f1d4e7.

Thanks @vaeth!

PR #344, opened 2018-06-18
Do not leak global variables REPLY and strategy

vaeth:
https://github.com/zsh-users/zsh-autosuggestions/issues/341


ericfreese:
Thanks for this! I have amended the commit to include the changes in the relevant `src/` files and pulled into `features/completion-suggestions` as 43a011026ffd9c7e9dba4cf022922a4877f1d4e7

Issue #343, opened 2018-06-18
disable if cursor is not at the end of the input line


vaeth:
If the cursor is not at the end of the input line, the suggestions are written at the end of the line anyway.
This is confusing (and has strange side effects if the suggestion should be accepted).

It would of course be nice if the suggestion could be inserted at the cursor position.
If this is technically not possible or too hard to implement (as I suspect), I suggest to temporarily switch off the suggestions if the cursor is not at the end of the input line.

vaeth:
But fish will show this suggestion at the end of the line?
The problem is that then accepting this suggestion gives you a nonsense command.

ericfreese:
In the case described on the PR, the buffer would be "dock" after typing the "o", the suggestion would add "er" to the end of it. Pressing `^E` twice accepts the "er" and leaves the buffer as "docker".

ericfreese:
Do you have an example that produces a nonsense command?

ericfreese:
Thanks for those examples. I hadn't considered those cases.  I've pushed another commit to #350 that moves the cursor to the end of the buffer in the completion pseudoterminal to ensure that completions are only generated at the end of the buffer. Let me know if that solves the issue for you.

I've come across another similar issue and I'm not sure how to solve it. Typing `ls "/"` results in a suggestion of `ls "/"ls "/bin"`. The issue seems to be that pressing tab after a quoted string moves your cursor to the end of the string in the quote and fetches completions there. Do you have any idea how to prevent this behavior? I may end up asking on the mailing list since I haven't come up with anything browsing through the docs for completion-related options and zstyles.

vaeth:
It works perfectly now.

I can confirm that pressing tab in a quoted string replaces the cursor, and I also did not find a zle setting to change this behaviour. Probably it is really best to ask on the ML.

ericfreese:
Ok I'm going to go ahead and close this ticket and open a new one for the issue described in my last comment.

PR #345, opened 2018-06-18
Offer suggestion only if cursor is at end of line

vaeth:
https://github.com/zsh-users/zsh-autosuggestions/issues/343


ericfreese:
Playing around with Fish shell a bit, it seems that Fish does give a suggestion even when the cursor is not at the end of the line. For example, if you've typed `dck` and insert an `o` between the `d` and the `c`, it will show a suggestion for `docker`.

Since Fish shell is the inspiration for this plugin, I think we should keep the behavior as-is.

PR #348, opened 2018-06-30
Fixes/no zpty

ericfreese:
Successor to #305. `zpty` is only required for `completion` suggestions, and will fail gracefully (not offer a suggestion) if `zpty` is unavailable.


src/strategies/completion.zsh
```diff
@@ -87,6 +87,8 @@ _zsh_autosuggest_capture_buffer() {
 }
 
 _zsh_autosuggest_capture_completion() {
+	zmodload -s zsh/zpty || return
```
okdana: The `-s` option to `zmodload` is only available in zsh 5.3+, so you'll be breaking compatibility with older versions of zsh (even if they have `zsh/zpty`) if you do this. Is that intended?

ericfreese: It is not intended! Thank you. I will remove

Issue #342, opened 2018-06-18
features/completion-suggestions differs from pressing TAB in subtle details


vaeth:
With features/completion-suggestions and strategy `completion`, the shown completion differs from that which is obtained by actually pressing TAB.

For instance, if /home is the only directory starting with /ho, then after `cd /ho` the suggested completion is `cd /home` while when pressing TAB would lead to 'cd /home/` (with a trailing slash). It would be much more convenient and less confusing to get the same result as with TAB.

vaeth:
It would probably also be helpful there would not be a longer suggestion shown than that which would be given by TAB (this even means __no__ suggestion in the non-unique case).

Indeed, it is very confusing if after typing `ls` one receives the suggestion `lsblk`, or even worse after typing `xma` one receives the suggestion `xmame.SDL` (when also `xmame.x11` and even `xmame` are binaries matching the same beginning), or if after typing `ls /usr/bin/` one receives the suggestion `ls /usr/bin/x11`.

Especially if one is used to typing TAB frequently (or using auto-fu.zsh), this gives the wrong intuition that `ls` is actually not a complete command, that `xmame.SDL` is the only binary starting with `xma` (in particular, there is no `xmame`), or that `x11` is the _only_  binary existing in `/usr/bin/`, respectively. (Note that x11 is not even the alphabetically first file there; in fact, the choice in the non-unique case seems to be random or depending on the disk hash function.)

ericfreese:
I think #350 addresses all of your concerns here. Take a look when you can.

vaeth:
The “cd /ho`me/`” is now working, and also “/usr/bin/` `” is fine.
However, there is still “ls`blk`” and “/usr/bin/xma`me.SDL`” (despite xmame exists).
The latter is perhaps correct, because pressing 2 times tab shows indeed xmame.SDL before xmame in the completion list for some strange reason.

ericfreese:
> However, there is still “lsblk” and “/usr/bin/xmame.SDL”

Strange- this is fixed for me. When I type `ls` I do not get a suggestion. After typing a `b`, I do get a suggestion for `lsblk`.

Pressing tab when the buffer has just `ls` gives these results:

```
ls          lsblk       lsinitcpio  lslocks     lsmem       lsns        lspcmcia    lsusb.py    
lsattr      lscpu       lsipc       lslogins    lsmod       lspci       lsusb  
```

Note `ls` is first, so there is no additional suggestion shown.

[...]

PR #350, opened 2018-07-01
Improve completion suggestions

ericfreese:
Just insert the first completion directly into the buffer and read the whole buffer from the zpty.

Should fix a couple issues reported in #342.

### DIFF
--- a/src/strategies/completion.zsh
+++ b/src/strategies/completion.zsh
@@ -17,6 +17,9 @@ _zsh_autosuggest_capture_completion_widget() {
 	local -a +h comppostfuncs
 	comppostfuncs=(_zsh_autosuggest_capture_postcompletion)
 
+	# Only capture completions at the end of the buffer
+	CURSOR=$#BUFFER
+
 	# Run the original widget wrapping `.complete-word` so we don't
 	# recursively try to fetch suggestions, since our pty is forked
 	# after autosuggestions is initialized.
--- a/src/strategies/completion.zsh
+++ b/src/strategies/completion.zsh
@@ -86,14 +86,16 @@ _zsh_autosuggest_strategy_completion() {
 		zpty -w $ZSH_AUTOSUGGEST_COMPLETIONS_PTY_NAME $'\t'
 	fi
 
-	# The completion result is surrounded by null bytes, so read the
-	# content between the first two null bytes.
-	zpty -r $ZSH_AUTOSUGGEST_COMPLETIONS_PTY_NAME line '*'$'\0''*'$'\0'
+	{
+		# The completion result is surrounded by null bytes, so read the
+		# content between the first two null bytes.
+		zpty -r $ZSH_AUTOSUGGEST_COMPLETIONS_PTY_NAME line '*'$'\0''*'$'\0'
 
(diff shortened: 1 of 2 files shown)


vaeth:
There is a problem with this patch which already happened sometimes with auto-fu.zsh:
Sometimes the input now hangs forever and can only be shut off with Ctrl-C.

In contrast to auto-fu.zsh, after a Ctrl-C in such a situation the shell can become unusable.
There occurs repeatedly the message
```
_zsh_autosuggest_strategy_completion:zpty:12: pty command name already used: zsh_autosuggest_completion_pty
```

vaeth:
I should add that this message appeared already sometimes without this patch, but I was not able to reproduce it like now. (And it did not make the shell unusable).

ericfreese:
> There is a problem with this patch which already happened sometimes with auto-fu.zsh:
Sometimes the input now hangs forever and can only be shut off with Ctrl-C.

Interesting. I was seeing this locally before, but it was fixed for me with this patch. I was seeing it most reliably when typing a space after `man`.

Have you found a case that reproduces the problem consistently? Can you provide more details?

ericfreese:
Also, I did just force push a small change that fixed a problem using autosuggest with zsh-syntax-highlighting after this patch. Maybe it would help with auto-fu as well. Can you try the most recent commit fc8c034?

vaeth:
After removing the line
```
zstyle ':completion:*' sort false
```
from my .zshrc due to #342 it turns out that the delay disappeared.
This is quite surprising, since I would have expected the opposite: That sorting takes time and not omitting sorting. I think I will report this on the zsh-dev mailing list.

ericfreese:
Thank you so much for your effort reproducing and bisecting your zshrc to find the offending zstyle. I've pushed a new commit 156ae5e that ensures that the zpty is always destroyed (even on ctrl-c) so that it can be recreated without issue the next time around. Please try the latest and let me know if it solves the issue for you.

As an aside, if you have completions that take a long time, you can use `ZSH_AUTOSUGGEST_USE_ASYNC=true` in your `.zshrc` to fetch suggestions asynchronously so they won't block your typing.

vaeth:
Indeed, https://github.com/zsh-users/zsh-autosuggestions/commit/156ae5ea69de4a68342593ee151ad288695093bf fixes the problem.

I had tried ZSH_AUTOSUGGEST_USE_ASYNC=true before, but it had not changed anything.
If I set it now (in the above commit), I get no suggestions at all.

ericfreese:
> If I set it now (in the above commit), I get no suggestions at all.

That's not good. Can you reproduce this using `zsh -f` and sourcing the plugin manually?

```
% zsh -f
%% autoload compinit && compinit; source path/to/zsh-autosuggestions.zsh; ZSH_AUTOSUGGEST_USE_ASYNC=true; ZSH_AUTOSUGGEST_STRATEGY=(completion)
```

vaeth:
Aha. Again it was a setting in my zshrc. This time the harmless looking
`NULLCMD=:`
is the culprit.

vaeth:
I have opened a new PR #352 for the latter.

Async mode now works great! Maybe it should become the default.

PR #352, opened 2018-07-02
Do not rely on implicit NULLCMD=cat

vaeth:


ericfreese:
Thank you! I have cherry-picked onto `develop` branch as 8ae0283

PR #355, opened 2018-07-03
Update INSTALL.md

amehmeto:
I lost myself in the internet abyss because of this imprecision. It might be helpful for the new noob trying to install that great tweak. :)


ericfreese:
Thanks! :+1:

PR #378, opened 2018-10-15
Don't overwrite config with default values

ericbn:
otherwise users are obliged to set the config values *after* sourcing the plugin. They're not able to do it before. Also, re-sourcing the plugin will reset the values to the defaults again.

See zimfw/zimfw#301

### DIFF
--- a/src/config.zsh
+++ b/src/config.zsh
@@ -6,17 +6,17 @@
 # Color to use when highlighting suggestion
 # Uses format of `region_highlight`
 # More info: http://zsh.sourceforge.net/Doc/Release/Zsh-Line-Editor.html#Zle-Widgets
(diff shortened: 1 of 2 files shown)


ericbn:
@ericfreese, good points. I'm working on that. The `: ${VAR:=default}` construct does not work for array values though. For example, the following creates a scalar value of `(history)` (instead of an array value containing just the element `history`):

    : ${ZSH_AUTOSUGGEST_STRATEGY:=(history)}

Which construct do you prefer for these cases?

ericfreese:
Thanks for your work on this @ericbn. I'm closing this PR in favor of #390. Please let me know what you think of that implementation.

PR #382, opened 2018-10-29
Feature: dynamic strategy switching

psprint:
This PR adds dynamic strategy switching. If first two characters are "@@", the strategy is set to `default`. If "%%", the strategy is set to `match_prev_cmd`. After 1 command, the situation goes back to normal – to user configured strategy. [Demonstration on Asciinema](https://asciinema.org/a/vVd6JBwOBkGxPScI5k5E8ZYp4).

I'm doing this PR for Hacktoberfest, I hope you find it useful (this is needed by the Hactoberfest's rules).

A new strategy would boost this switching's value, I'm thinking of it, I'm aware how hard it is.

### DIFF
--- a/zsh-autosuggestions.zsh
+++ b/zsh-autosuggestions.zsh
@@ -49,6 +49,25 @@ ZSH_AUTOSUGGEST_ORIGINAL_WIDGET_PREFIX=autosuggest-orig-
 
 ZSH_AUTOSUGGEST_STRATEGY=default
 
+# Holds currently recognized trigger – two first characters
+# of $BUFFER can be special, see the next parameter
+ZSH_AUTOSUGGEST_CURRENT_TRIGGER=""
+
+# Dynamically change the strategy - first two characters
+# of $BUFFER can be recognized as special: if they occur
+# in the map below - and if so then the strategy will be
+# changed for current single command
+typeset -gA ZSH_AUTOSUGGEST_DYNAMIC_STRATEGY_TRIGGER_MAP
+ZSH_AUTOSUGGEST_DYNAMIC_STRATEGY_TRIGGER_MAP=(
+	"@@" default
+	"%%" match_prev_cmd
+)
+
+# If the possible trigger will match above hash,
+# then a custom, dynamic strategy will be selected
+# and stored in this scalar; empty if no match.
+ZSH_AUTOSUGGEST_DYNAMIC_STRATEGY=""
+
 # Widgets that clear the suggestion
 ZSH_AUTOSUGGEST_CLEAR_WIDGETS=(
 	history-search-forward
@@ -211,6 +230,9 @@ _zsh_autosuggest_bind_widget() {
 
 # Map all configured widgets to the right autosuggest widgets
 _zsh_autosuggest_bind_widgets() {
+        # Clear dynamic-strategy
+        ZSH_AUTOSUGGEST_CURRENT_TRIGGER="" ZSH_AUTOSUGGEST_DYNAMIC_STRATEGY=""
+
 	local widget
 	local ignore_widgets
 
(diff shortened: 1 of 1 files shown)


psprint:
Having an idea-spree, the strategies are:
- when no line-continuation is found in `$history`, then fallback to using a dictionary, which might contain: long-built, precious commands; with this PR, one can also switch to dictionary directly, e.g. via "&&" on beginning of the command line,
- a dictionary of lines from a set of files – the autoload functions shipped with Zsh, in /usr/local/share/zsh/{version}/functions; I think this could serve as a way to try-out new things; also assign the two-character switching-trigger,
- simple strategy – complete Git aliases from `~/.gitconfig`; I mean completion of alias body, contents, not the alias name; not sure about a use case, maybe: non-distracting (no interruption with `vim ~/.gitconfig`) way of finding how is an alias built; also: it might be sensible that trying to get familiar with a Git alias is better by first looking on its contents, not name.

ericfreese:
I think I'm not interested in pulling this functionality into the plugin itself, but I think this code could make a decent plugin by itself (run arbitrary code when certain triggers are entered into the buffer), and I would be happy to consider a PR that made this plugin easier to integrate with the new plugin (things like adding functionality to recreate the zpty though not actually that since the pry is going away on develop branch). But I'm not interested in adding the trigger logic and config options.

I would re-open this and reconsider though if there's a significant number of people interested in adding this functionality.

psprint:
I've failed with coming up with an additional reasonable strategy, forgot about it, after doing some brain storm. Maybe if there will be a reasonable strategy this PR could be reconsidered as well?

ericfreese:
> Maybe if there will be a reasonable strategy this PR could be reconsidered as well?

I think even with more strategies, this functionality would be better extracted to a separate plugin. I'm picturing a plugin with a goal of allowing users to run arbitrary commands when certain characters are entered at the beginning of the buffer. Then a user could configure _that_ plugin to, when they enter certain keystrokes, switch to certain zsh-autosuggestions strategies.

I think adding more strategies should be a separate PR, but I'm honestly quite hesitant to add more strategies since I think they're so personal and belong more or less in an individual's configuration... I'm wondering if it would be helpful to have a separate "contrib" repo for strategies that people can copy/paste/install into their zshrc.

psprint:
Hmm indeed thinking now to switch the strategy the other plugin would have to alter the single global parameter holding the strategy to use. However, in async-mode, it would also need to restart zpty. But this could belong to some general integration to do with the arbitrary-action plugin.

I thought that a new strategy is really hard to come up with, as there are apparently not much hmm... degrees of freedom in the task of: "searching a history for most recent match". Hard to add or change something within this. But you say people actually have some alternative strategies but they're really custom? I've only saw the history[@]->sqlite plugin that made a strategy working on database.

ericfreese:
> However, in async-mode, it would also need to restart zpty.

This is no longer true on `develop` thanks to your `< ()` trick :smile:

> But this could belong to some general integration to do with the arbitrary-action plugin.

Yes, that's what I'm thinking. If we can add some code to this repo to make it easier for other plugins to interact with it, I would be open to that (as long as it doesn't make things too messy).

> But you say people actually have some alternative strategies but they're really custom? I've only saw the history[@]->sqlite plugin that made a strategy working on database.

That's the one for me that immediately comes to mind, but there have been many other suggestions for strategies like the ones you brought up above. I'm thinking strategies like those would be better off implemented in individual's configs.

psprint:
> This is no longer true on `develop` thanks to your `< ()` trick 😄

BTW. Alexandros Kozák did a few observations of the `<()` method: https://github.com/agkozak/agkozak-zsh-prompt/pull/11

PR #386, opened 2018-11-04
Add history cycling widgets

m-lima:
This change adds two widgets to the project. They work as a pair and allow to cycle up and down the suggestions without having to commit to a suggestion.

This works similarly to `zsh`'s <kbd>up</kbd> and <kbd>down</kbd>, but without filling the `BUFFER`. The user can then call `autosuggest_execute` or `autosuggest_accept`, or any other `zsh_autosuggestions` widget for that matter.
The cycling gets capped at the top and bottom of the history matches.


m-lima:
Closing because of conflicts with `develop`

m-lima:
Re opening to keep it visible.
Will keep on working on the `rebase` for now

m-lima:
Rebase done

PR #387, opened 2018-11-04
Feature: Allow cycling throw history in the autocompletion

m-lima:
This change adds two widgets to the project. They work as a pair and allow to cycle up and down the suggestions without having to commit to a suggestion.

This works similarly to `zsh`'s <kbd>up</kbd> and <kbd>down</kbd>, but without filling the `BUFFER`. The user can then call `autosuggest_execute` or `autosuggest_accept`, or any other `zsh_autosuggestions` widget for that matter.
The cycling gets capped at the top and bottom of the history matches.


ericfreese:
Going to go ahead and close this since it's not complete and hasn't had any activity for a long time. Feel free to pick this up again at some time in the future and open a new PR if you like.

PR #390, opened 2018-11-10
Features/no overwrite config

ericfreese:
Builds on #378 but:

- uses `: ${var=thing}` instead of `: ${var:=thing}` for scalars and `[[ -z "${var+x}" ]]` instead of `(( ! ${#var} ))` for arrays so that we only set the config params if they are in fact unset, and not if they are just null (e.g. empty string, empty array).
- updates specs to set options _before_ the plugin is sourced instead of after.

@ericbn What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ If you invoke the `forward-word` widget, it will partially accept the suggestion
 
 ## Configuration
 
-You may want to override the default global config variables after sourcing the plugin. Default values of these variables can be found [here](src/config.zsh).
+You may want to override the default global config variables. Default values of these variables can be found [here](src/config.zsh).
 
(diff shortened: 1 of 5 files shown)


ericbn:
Hi @ericfreese. I would go with checking if the arrays are not set with `! ${+VAR}` as in

    (( ! ${+ZSH_AUTOSUGGEST_STRATEGY} )) && ZSH_AUTOSUGGEST_STRATEGY=(history)

This will keep the same behavior of `${VAR=default}` used for the scalar variables. From zshexpn(1):

    ${+name}
           If name is the name of a set parameter `1' is substituted, otherwise `0' is substituted.

    ${name=word}
    ...
           In the first form, if name is unset then set it to word;

So both

    : ${VAR=default}
    (( ! ${+VAR} )) && VAR=default

mean exactly the same thing: if `VAR` is unset, then set it to `default`.

ericfreese:
Updated. Thanks @ericbn!

ericbn:
Cool! Looking forward to seeing this change in the master branch.

Issue #411, opened 2019-03-19
rebinding widgets slows down prompt display


leotaku:
This is related to my previous issue at #346, but the functionality surrounding the problem has since been reworked, which is why I chose to close said issue and open a new one.

### My problem:
zsh-autosuggestions is setup in a way that it rebinds all widgets on every precmd call. [here](https://github.com/zsh-users/zsh-autosuggestions/blob/master/zsh-autosuggestions.zsh#L734) 

Unfortunately, for me at least, this seems to cause a noticeable slowdown when displaying a new prompt.

### Possible fix
I don't think widget definitions change all that much during "normal" shell operation. So it could be a good idea for most use-cases to simply remove the hook or at least make it configurable.

romkatv:
FYI: There is a PR that speeds this code up by 20%. Not much, but for the size of the PR it's not terrible. The PR has failed to elicit any reaction from the project owner though.

ericfreese:
Put up PR #418 to add `ZSH_AUTOSUGGEST_MANUAL_REBIND` config option to disable automatic re-binding. Give it a shot and I'll merge it if there's no objections. Made my prompts about twice as fast according to @romkatv's prompt benchmark :+1:

PR #418, opened 2019-04-10
Allow disabling of automatic widget re-binding 

ericfreese:
Addresses github #411

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,10 @@ This can be useful when pasting large amount of text in the terminal, to avoid t
 
 As of `v0.4.0`, suggestions can be fetched asynchronously. To enable this behavior, set the `ZSH_AUTOSUGGEST_USE_ASYNC` variable (it can be set to anything).
 
+### Disabling automatic widget re-binding
+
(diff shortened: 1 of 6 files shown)


romkatv:
Most plugins rebind widgets only during sourcing. To ensure that zsh-autosuggestions rebinds widgets after everyone else, you could hook `precmd`, run `_zsh_autosuggest_start` from it, and then unhook `precmd` if `ZSH_AUTOSUGGEST_MANUAL_REBIND` is set. This would be more robust than the current code because it will work regardless of the order in which different plugins are sources. The only case where `ZSH_AUTOSUGGEST_MANUAL_REBIND` wouldn't work (compared to the current code that rebinds everything on every `precmd`) is when widgets are defined or rebound after the first prompt is rendered. This limitation is reasonable and easy to understand.

ericfreese:
> you could hook precmd, run _zsh_autosuggest_start from it, and then unhook precmd if ZSH_AUTOSUGGEST_MANUAL_REBIND is set

Cool, this is essentially what the PR does :smile:

romkatv:
> Cool, this is essentially what the PR does

:man_facepalming:

LGTM!

leotaku:
Great, Thank you! This seems like it's the best option for most users. Although I will probably still remove the `precmd` hook and run `_zsh_autosuggest_start` in my configuration, just to avoid the flicker produced by the first `precmd`.

PR #420, opened 2019-04-15
Fixes/completion newline issues

ericfreese:


src/strategies/completion.zsh
```diff
@@ -99,8 +108,10 @@ _zsh_autosuggest_strategy_completion() {
 		# content between the first two null bytes.
 		zpty -r $ZSH_AUTOSUGGEST_COMPLETIONS_PTY_NAME line '*'$'\0''*'$'\0'
 
-		# On older versions of zsh, we sometimes get extra bytes after the
-		# second null byte, so trim those off the end
+		# Extract the suggestion from between the null bytes.  On older
+		# versions of zsh (older than 5.3), we sometimes get extra bytes after
+		# the second null byte, so trim those off the end.
+		# See http://www.zsh.org/mla/workers/2015/msg03290.html
 		suggestion="${${${(M)line:#*$'\0'*$'\0'*}#*$'\0'}%%$'\0'*}"
```
ericfreese: Can we use `${${(0)line}[2]}` here?

ericfreese: Looked into this. Answer appears to be no because the string before the first null byte is sometimes empty and is omitted from the resulting array. So the index needed to access the string between the null bytes is either 1 or 2 depending on whether this condition is true. Found and implemented a simpler method though.

PR #430, opened 2019-05-18
Update README.md

DanTheColoradan:
Clarify what fg stands for.

https://stackoverflow.com/questions/47310537/how-to-change-zsh-autosuggestions-color


ericfreese:
Hi @DanTheColoradan, thank you for this! There is actually quite a bit more functionality available than just the 256-color palette. I've gone ahead and made a separate PR #434 to improve the documentation. Take a look and let me know what you think.

Issue #431, opened 2019-05-25
Highlight not applied when zle-line-init modifies BUFFER


romkatv:
Short version. zsh-syntax-highlighting wraps `zle-line-init` but zsh-autosuggestions doesn't. Thus, if `zle-line-init` modifies `BUFFER`, zsh-autosuggestions won't apply its highlighting to `POSTDISPLAY`, which will make it appear as if `POSTDISPLAY` was a part of `BUFFER`.

Why am I modifying `BUFFER` in `zle-line-init`? I'm using something like https://github.com/robbyrussell/oh-my-zsh/tree/master/plugins/dirhistory but my version doesn't clear `BUFFER` when I'm changing directories. Here's a simplified example that allows me to execute `cd ..` when pressing `Ctrl-U` without clearing `BUFFER`.

```zsh
function cd-up() {
  cd ..
  saved_buffer=$BUFFER
  saved_cursor=$CURSOR
  BUFFER=
  zle accept-line
}

function restore-buffer() {
  BUFFER=$saved_buffer
  CURSOR=$saved_cursor
  saved_buffer=
  saved_cursor=1
}

zle -N cd-up
zle -N restore-buffer

autoload -Uz add-zle-hook-widget
add-zle-hook-widget line-init restore-buffer

bindkey "^U" cd-up
```

Note that zsh-autosuggestions wraps `cd-up` widget and adds highlighting after it modifies `BUFFER` but unfortunately this is not enough. The wrapper code for `zle-line-init` in zsh-syntax-highlighting runs afterwards and resets the highlighting.

Wrapping `zle-line-init` in zsh-autosuggestions fixes this issue but unfortunately there is no option to do it. Currently I'm using this hack:

```zsh
functions[orig_zsh_autosuggest_bind_widgets]=$functions[_zsh_autosuggest_bind_widgets]
function _zsh_autosuggest_bind_widgets() {
  orig_zsh_autosuggest_bind_widgets
  _zsh_autosuggest_bind_widget zle-line-init modify
}
```

You can see how it looks without and with the hack in https://asciinema.org/a/iU6zszJucHfbqVyr6sNSMuSDT?t=21. This records the following:

```zsh
docker run -e LANG=C.UTF-8 -e LC_ALL=C.UTF-8 -e TERM=$TERM -it --rm ubuntu bash -uexc '
  cd
  apt update && apt install -y zsh git
  git clone https://github.com/zsh-users/zsh-syntax-highlighting.git
[...]

PR #432, opened 2019-05-27
Wrap zle-line-init to support fetching suggestions on line init

ericfreese:
Use case suggested by @romkatv uses zle-line-init to restore buffer after running a widget to cd up one level (GitHub #431).

As far as I can tell, the ignoring of zle-line-* was added in commit 9788c2e to support some deprecation warnings that were removed some time ago.

The pattern was then widened in commit 0c940e7 to zle-* to fix problems encountered when wrapping zle-isearch-update.

This commit removes the hard coded ignore of all zle-* widgets and puts zle-isearch-update and zle-keymap-select into the default list of widgets to be ignored. It leaves zle-line-init and zle-line-finish not ignored so that suggestions will be fetched after they are run.

### DIFF
--- a/spec/line_init_spec.rb
+++ b/spec/line_init_spec.rb
@@ -1,0 +1,17 @@
+context 'with zle-line-init unignored' do
+  let(:after_sourcing) do
+    -> do
+      session.
+        run_command('setopt extendedglob').
+        run_command('ZSH_AUTOSUGGEST_IGNORE_WIDGETS=(${(@)ZSH_AUTOSUGGEST_IGNORE_WIDGETS:#zle-\*} zle-\^line-init)').
+        run_command('zle-line-init() { BUFFER="echo" }')
+    end
+  end
+
+  it 'should fetch a suggestion on each line initialization' do
+    with_history('echo foo') do
+      session.run_command('zle -N zle-line-init')
+      wait_for { session.content }.to end_with('echo foo')
+    end
+  end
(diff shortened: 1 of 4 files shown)


romkatv:
Thanks for sending the PR.

After a bit of testing I bumped into a corner case that gives an error. Type this into `zsh -df`:

```zsh
ZSH_AUTOSUGGEST_MANUAL_REBIND=1
autoload -Uz add-zle-hook-widget
autoload -Uz add-zsh-hook
function f g() {}
add-zle-hook-widget line-finish f
source zsh-autosuggestions.zsh
function h() { add-zle-hook-widget line-finish g }
add-zsh-hook precmd h
```

Then hit `<enter>`. You'll see "maximum nested function level reached" error.

I've added workarounds in my configs but it's possible that others will have this problem. It might be safer to add `zle-line-finish` and `zle-line-init` to the default `ZSH_AUTOSUGGEST_IGNORE_WIDGETS`, so that users who don't override the defaults get the same behavior as before.

Or maybe you could fix the underlying issue with the mutual recursion of wrappers within `add-zle-hook-widget` and zsh-autosuggestions. I didn't look into it.

ericfreese:
Thanks @romkatv :+1: I'll look briefly into fixing the issue you've brought up, but will likely just go ahead and add zle-line-init and zle-line-finish to the ignore widgets as you've suggested.

Issue #437, opened 2019-06-14
Breaks copy-earlier-word


henrebotha:
# Steps taken

1. Add to `~/.zshrc`: ```autoload -Uz copy-earlier-word; zle -N copy-earlier-word; bindkey '^N' copy-earlier-word```.
2. Source `~/.zshrc`.
3. Type `foo bar baz ` (note the trailing space).
4. Press <kbd>^</kbd><kbd>N</kbd> multiple times.

# Observed result

The words produced are `baz`, `baz`, `bar` in that order, repeating with further keypresses.

# Expected result

The words produced are `baz`, `bar`, `foo`, and an empty string, in that order, with no repetitions upon further keypresses.

# Environment

I have observed this in Zsh 5.5.1 and 5.7.1, both using Tmux and not using Tmux, under xfce4-terminal 0.8.7.4 on Xubuntu 18.10. I have observed this with zsh-autosuggestions at `70f36c0` and `733abd4`.

I have tried multiple versions of Zsh config (e.g. with and without vi-style bindings), and observed the error in all of them.

ericfreese:
Fixed on develop, see #439

PR #438, opened 2019-06-14
Clear the suggestion on copy-earlier-word

henrebotha:
The test isn't correct, I can't actually get it to fail. Hoping for some input here.


henrebotha:
This aims to fix #437 .

ericfreese:
Thanks! It looks like async mode has to be enabled for `copy-earlier-word` to fail. See #319 for a fix to a very similar issue. Any widget that relies on $LASTWIDGET will likely break in async mode, since the `autosuggest-suggest` widget is called asynchronously when the suggestion is ready, overwriting the previous value of $LASTWIDGET that the other widget is expecting.

Issue #442, opened 2019-06-17
Parse error since last update


n3f4s:
Hi,

I just updated zsh-autosuggestions via `pacman` and since then I have a parse error:
```
/usr/share/zsh/plugins/zsh-autosuggestions/zsh-autosuggestions.zsh:9: defining function based on alias `_complete'
/usr/share/zsh/plugins/zsh-autosuggestions/zsh-autosuggestions.zsh:571: parse error near `()'

```

update log from `pacman`:
```
[2019-06-17 03:17] [ALPM] upgraded zsh-autosuggestions (0.5.2-1 -> 0.6.0-1)
```

farzadmf:
I have the same issue

widgetii:
Me too

ericfreese:
Hi, thanks for reporting. I'm not immediately able to reproduce, so I'll need some more info from either of you.

1. Can you reproduce this without running other commands in your zshrc?

```sh
% zsh -df
% source /usr/share/zsh/plugins/zsh-autosuggestions/zsh-autosuggestions.zsh # or whatever the path is
```

If not, can you please comment out sections of your init files (.zshrc, etc.) until we narrow down the root cause of the issue?

2. What version of zsh are you running?

```sh
% echo $ZSH_VERSION
```

3. What zsh options do you have enabled?

```sh
% setopt
```

n3f4s:
1. When running
```bash
% zsh -df
% source /usr/share/zsh/plugins/zsh-autosuggestions/zsh-autosuggestions.zsh
```
I have no error message

2. `zsh 5.7.1 (x86_64-pc-linux-gnu)`

3.
```bash
% setopt
alwaystoend
autocd
autopushd
nobeep
completeinword
correctall
extendedglob
extendedhistory
noflowcontrol
histexpiredupsfirst
histignoredups
histignorespace
histverify
incappendhistory
interactive
interactivecomments
longlistjobs
monitor
promptsubst
pushdignoredups
pushdminus
sharehistory
shinstdin
zle
```

After commenting/de-commenting it would seems that the issue come from this part (especially the call to `bash_source`):
```bash
__completion_loader() {
	compdef "$2=$1"
}
bash_source() {
  alias shopt=':'
  alias _expand=_bash_expand
  alias _complete=_bash_comp
  emulate -L sh
  setopt kshglob noshglob braceexpand
  local name=$(basename $1)
  alias _completion_loader=__completion_loader $name

  source "$@"
}
[...]

PR #445, opened 2019-06-18
Added 'function' keyword

pkobielak:
After installation of the plugin I've got following errors:
```
/home/user/.oh-my-zsh/custom/plugins/zsh-autosuggestions/zsh-autosuggestions.zsh:9: defining function based on alias `_complete'
/home/user/.oh-my-zsh/custom/plugins/zsh-autosuggestions/zsh-autosuggestions.zsh:571: parse error near `()'
```
It turned out that there is a lack of 'function' keyword near '_complete' function.
Here's a simple fix.


ericfreese:
Thanks @pkobielak. This has actually already been merged to develop branch. See #442 and #445.

PR #476, opened 2019-10-10
(README) Link to 256 color chart

gregrickaby:
Easy to see and choose a color, and then copy it over from the chart.


ericfreese:
Thanks!

Issue #463, opened 2019-07-29
Add a way to disable *completion* strategy in certain cases (similar to history-ignore)


vaeth:
Although completion strategy works now like a charm, there can be a problem:

I have a command at work whose completion function is very complex and not maintained by me. This completion calls a program which needs quite some resources (memory, filehandles - I don't know). When tipping too quickly, too many instances of said program are started which can lead to a total crash of the shell or even of the machine.

Therefore, I would like to disable completion auto-suggestion for that command.
Of course, I could disable completion for that command completely, but actually I would like to keep the "manual" completion (as mentioned, it is very complex).

Would it be possible to add options for e.g. a pattern/regexp such that when the line and/or command matches said pattern the completion auto-suggestion gets disabled?

The feature could be similar to history-ignore but does not refer to entries in the history file but the current command line.

ericfreese:
Great idea. I'll look into implementing this soon :+1:

ericfreese:
@vaeth Apologies for the delay on this. Will you please give branch `features/completion-ignore` (PR https://github.com/zsh-users/zsh-autosuggestions/pull/487) a try and see if it meets your needs? Thank you!

vaeth:
Works great! Thank you very much!

PR #487, opened 2019-12-15
Allow skipping completion suggestions when buffer matches a pattern

ericfreese:
Set ZSH_AUTOSUGGEST_COMPLETION_IGNORE to a glob pattern to have the completion suggestion strategy never make suggestions when the buffer matches the pattern.

This can be helpful when some completion routines you have are particularly expensive and you want to prevent them from running automatically on every keystroke.

See GitHub issue #463.


mplattner:
Great, I was looking for this feature. Thanks, Eric.

Issue #470, opened 2019-09-15
Prompt is slow after multiple "source ~/.zshrc"


silverneko:
### Describe the bug
zsh prompt becomes very slow after multiple times of "source ~/.zshrc". I suspect it's the widgets of "_zsh_autosuggest_bound_#n_*" slowing zsh down, because after every time I do "source ~/.zshrc" the number of these widgets increases. For example when a new shell starts only widgets of "_zsh_autosuggest_bound_1_*" are registered, and after a second "source ~/.zshrc" widgets of "_zsh_autosuggest_bound_2_*" are added.

I use oh-my-zsh to load this plugin.

### To Reproduce
Do this for 10 to 20 times:
```
$ source ~/.zshrc
```

### Desktop
 - OS + distribution: debian
 - Zsh version: 5.7
 - Plugin version: 43f3bc40

ericfreese:
I could see this being an issue but will probably be low priority for me to fix since the issue is avoidable by not re-sourcing zshrc many times.

silverneko:
You're right! It's not really a issue for normal users. I only encounter this issue because I'm testing with different .zshrc configurations thus sourcing .zshrc multiple times.

z0rc:
Sourcing `.zshrc` isn't optimal in many ways and will break more than z-asug. Instead it's better to use `exec zsh` to completely reinit from clean state.

NightMachinary:
@silverneko How do you check the list of all widgets?

silverneko:
`zle -l` or if you have auto completion then `_<tab>`

PR #495, opened 2020-01-06
Update INSTALL.md

sagivo:
add option to reload the shell without exiting.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -42,7 +42,7 @@
     plugins=(zsh-autosuggestions)
     ```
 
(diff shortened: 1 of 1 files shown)


ericfreese:
Thanks for this, but I don't actually recommend sourcing `~/.zshrc` multiple times in the same session. [`exec zsh` seems to be a better option](https://github.com/zsh-users/zsh-autosuggestions/issues/470#issuecomment-532200982), but neither of them are as fool-proof as starting a new session from scratch, so I think that is what I would like to continue suggesting.

Issue #481, opened 2019-11-28
background initialization


horosgrisa:
### Describe the bug

I load this plugin in the background, after `precmd` hook.
This plugin runs `_zsh_autosuggest_start`  after `precmd`.
So, first prompt will be without autosuggestions. 

### To Reproduce
```sh
TMOUT=1

TRAPALRM() {
  source $PATH_TO_PLUGIN/zsh-autosuggestions.plugin.zsh
}
```

### Expected behavior

After initialization, the plugin must run `_zsh_autosuggest_bind_widgets` directly

ericfreese:
Hi, thanks for reporting this. Please try branch `features/background-initialization` from PR #506 and let me know if it solves your issue.

horosgrisa:
@ericfreese Hi, thank you very much. It works!

horosgrisa:
@ericfreese Hi, why was this branch not merged?

ericfreese:
Simple answer is I just haven't had time to get to it. I'll try to get it merged soon 👍

PR #506, opened 2020-01-27
Allow loading the plugin in the background

ericfreese:
If zle is already active (plugin has been loaded in the background), just start the widgets and set up the automatic re-binding (if manual rebind has not been specified).

See GitHub issue #481

### DIFF
--- a/src/strategies/completion.zsh
+++ b/src/strategies/completion.zsh
@@ -45,8 +45,6 @@ _zsh_autosuggest_capture_completion_widget() {
 zle -N autosuggest-capture-completion _zsh_autosuggest_capture_completion_widget
 
 _zsh_autosuggest_capture_setup() {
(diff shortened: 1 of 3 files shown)


src/strategies/completion.zsh
```diff
@@ -45,8 +45,6 @@ _zsh_autosuggest_capture_completion_widget() {
 zle -N autosuggest-capture-completion _zsh_autosuggest_capture_completion_widget
 
 _zsh_autosuggest_capture_setup() {
-	autoload -Uz is-at-least
```
ericfreese, line 48: There is now one more is-at-least autoload that could possibly be cleaned up

https://github.com/zsh-users/zsh-autosuggestions/blob/aa05920a4aa5a54c2aadbafe66771d1874671e11/src/async.zsh#L47-L47

Issue #483, opened 2019-12-02
Conflict with `zsh-users/zsh-syntax-highlighting`


elony314:
### Describe the bug
There are conflicted between `zsh-users/zsh-syntax-highlighting` and `zsh-users/zsh-autosuggestions`. With both plugins activated, if one continuously uses `ZSH_AUTOSUGGEST_PARTIAL_ACCEPT_WIDGETS`, the last word of the suggestion will be duplicated.

### To Reproduce
```sh
% zsh -df
% source path/to/zsh-autosuggestions.zsh
% source path/to/zsh-syntax-highlighting.plugin.zsh
% export WORDCHARS='*?_[]~=&;!#$%^(){}'
% cd Test1/Test2
```
Now, input `cd` and continuously use `ZSH_AUTOSUGGEST_PARTIAL_ACCEPT_WIDGETS` until the end of the line. Instead of get the complete `Test1/Test2`, the `Test2` part will be duplicated.

### Expected behavior
```sh
cd Test1/Test2
```

### Actual behavior
```
cd Test1/Test2Test2
```

### Desktop
 - OS + distribution: macOS 10.15.1
 - Zsh version: 5.7.1
 - Plugin version: d43c309f888153d6c46d8b6a3a0186f4148680fd

elony314:
I have also submitted this bug [here](https://github.com/zsh-users/zsh-syntax-highlighting/issues/647)

ericfreese:
I'm not able to reproduce this either. What are you using to trigger `ZSH_AUTOSUGGEST_PARTIAL_ACCEPT_WIDGETS`? What version of z-sy-h?

ericfreese:
@elony314 I was able to reproduce this and put up PR #507 to fix. Give that branch (`fixes/partial-accept-duplicate-word`) a shot and let me know if it looks good to you.

weslley39:
Hi folks, first of all, amazing work that you guys are doing.

I came across this issue, but I don't know for sure If is the same.

I record a gif to show what is happening.

![ezgif-7-0841c4d7145c](https://user-images.githubusercontent.com/5286092/74858855-c2c77100-5324-11ea-86fa-7783cdfad44c.gif)


What I did, typed `cd`, pasted something and then pressed the right arrow key.
The suggestion was duplicated because was the last command that I used.

nathanshelly:
@weslley39 I think the original poster was describing the issue from this gif:

[...]

Issue #126, opened 2016-02-24
Segfault after sourcing zshrc again - conflict with zsh-autosuggestions even when sourced in right order.


torsava:
Hi, this issue is similar to issue #96:
After opening `zsh` and sourcing `.zshrc`, zsh exits with "`zsh: segmentation fault  zsh`":

```
% zsh
% . ~/.zshrc
zsh: segmentation fault  zsh
```

Issue #96 was solved by sourcing `zsh-autosuggestions` before `zsh-syntax-highlighting`, but that solution no longer works. Zsh segfaults no matter their order.
Here's the .zshrc file:

```
source ~/bin/zsh/antigen/antigen.zsh
antigen bundle tarruda/zsh-autosuggestions
antigen bundle zsh-users/zsh-syntax-highlighting
antigen apply
```

ericfreese:
This is only when sourcing the .zshrc for the second time correct?

I.e. You don't get a segfault if you do this?

```
% zsh -f
% . ~/.zshrc
```

tomparys:
Correct.

ericfreese:
I've been experimenting a bit.

Some test cases:

A) Sourcing only this plugin multiple times **(no segfault)**

```shell
% zsh -f
%% source ~/Code/zsh-autosuggestions/zsh-autosuggestions.zsh
%% source ~/Code/zsh-autosuggestions/zsh-autosuggestions.zsh
%% # no segfaults here...
```

B) Sourcing only the `zsh-syntax-highlighting` plugin multiple times **(no segfault)**

```shell
% git -C ~/Code/zsh-syntax-highlighting rev-parse --short HEAD
7044c19
% zsh -f
%% source ~/Code/zsh-syntax-highlighting/zsh-syntax-highlighting.zsh
%% source ~/Code/zsh-syntax-highlighting/zsh-syntax-highlighting.zsh
%% # no segfaults here...
```

C) Sourcing `zsh-syntax-highlighting` multiple times, then this plugin multiple times **(no segfault)**

```shell
% zsh -f
%% source ~/Code/zsh-syntax-highlighting/zsh-syntax-highlighting.zsh
%% source ~/Code/zsh-syntax-highlighting/zsh-syntax-highlighting.zsh
%% source ~/Code/zsh-autosuggestions/zsh-autosuggestions.zsh
%% source ~/Code/zsh-autosuggestions/zsh-autosuggestions.zsh
%% # no segfaults here...
```

[...]

PR #507, opened 2020-01-30
Fix issue where partial accept duplicates last word

ericfreese:
When z-sy-h is enabled after autosuggestion widgets have already been bound, partial accepting the last part of a suggestion will result in that string being duplicated.

I was able to reproduce using the following versions of the two plugins:
- z-sy-h: dde84e1b25f059a298ce3189cddfd0778f998df3
- z-asug: ae315ded4dba10685dbbafbfa2ff3c1aefeb490d

and running the following commands interactively one after another:

```
zsh -df
% source path/to/zsh-autosuggestions.zsh
% source path/to/zsh-syntax-highlighting.zsh
% bindkey -e
% bindkey <alt-f>         # shows 'bindkey -e-e'
```

The order of the `source` statements matters and the issue cannot be reproduced if the two source statements are executed together on one line.

The problem is very similar to this one:
  https://github.com/zsh-users/zsh-autosuggestions/issues/126#issuecomment-217121315

See GitHub issue #483


ericfreese:
CI is being flaky since the async by default PR went in. It looks like `with_history` spec helper is not working right. Seems to be entering the history commands only partially.

tojin12341za:
Duplicate of #

festival-up (approved):

stevenxxiu:
:+1: This fixes the issue for me, when it occurs when I use a custom widget:

```zsh
$ zsh -df
$ source /usr/share/zsh/plugins/zsh-autosuggestions/zsh-autosuggestions.zsh
$ function key-cright() {
  zle vi-forward-word -- $@
}
zle -N key-cright
bindkey $'\e[1;5C' key-cright
ZSH_AUTOSUGGEST_PARTIAL_ACCEPT_WIDGETS+=(key-cright)
```

I'm not sure why `unset POSTDISPLAY` fixes the issue. Is it because `POSTDISPLAY` is used in `_zsh_autosuggest_highlight_apply()`?

arisu2718:
Thanks, your fix works for me. Didn't even think I could resolve this issue so fast :D

PR #518, opened 2020-03-22
Add additional missing installation instructions for Manjaro/Arch Linux

BlueskyFR:
I added missing instructions for installation on Manjaro and Arch Linux. These instructions were once present in this file, I don't know why they got removed.
I searched for a while for this missing step so I think everyone should be aware of it, otherwise the installation just doesn't work.


mindyourlifeguide:
For example, I install it on **Arch Linux** to myself like this and it works:

```
# Install zsh-autosuggestions if it’s not present
if [[ ! -d ~/.oh-my-zsh/custom/plugins/zsh-autosuggestions ]]; then
  sudo git clone https://github.com/zsh-users/zsh-autosuggestions ~/.oh-my-zsh/custom/plugins/zsh-autosuggestions
fi
```

I just added this code to the beginning of _.zshrc_ and that’s it.

Issue #363, opened 2018-07-30
Autosuggest eats buffer by ctrl+w


jsirex:
Given command: `echo first second`
I want to cut two words `first second` by typing CTRL+WW . 

But then when I try to paste it with CTRL+Y I'm getting only one word: `first`.

I'm using oh-my-zsh:
```
plugins=(aws backup git extract knife knife_ssh mercurial mvn vagrant bundler gem rake rvm thor debian sudo kitchen docker-compose emacs berkshelf power-save terraform systemd zsh-autosuggestions)
```

ericfreese:
I pushed a failing spec to `fixes/kill-multiple-words`

```
multiple words killed with `backward-kill-word`
  can be yanked back with `yank` (FAILED - 1)

Failures:

  1) multiple words killed with `backward-kill-word` can be yanked back with `yank`
     Failure/Error: wait_for { session.content }.to eq('echo first second')
     
       expected: "echo first second"
            got: "echo first"
     
       (compared using ==)
     # /usr/local/bundle/gems/rspec-wait-0.0.9/lib/rspec/wait/handler.rb:13:in `block (2 levels) in handle_matcher'
     # /usr/local/bundle/gems/rspec-wait-0.0.9/lib/rspec/wait/handler.rb:10:in `loop'
     # /usr/local/bundle/gems/rspec-wait-0.0.9/lib/rspec/wait/handler.rb:10:in `block in handle_matcher'
     # /usr/local/bundle/gems/rspec-wait-0.0.9/lib/rspec/wait/handler.rb:9:in `handle_matcher'
     # /usr/local/bundle/gems/rspec-wait-0.0.9/lib/rspec/wait/target.rb:30:in `block in to'
     # /usr/local/bundle/gems/rspec-wait-0.0.9/lib/rspec/wait/target.rb:44:in `block in with_wait'
     # /usr/local/bundle/gems/rspec-wait-0.0.9/lib/rspec/wait.rb:28:in `with_wait'
     # /usr/local/bundle/gems/rspec-wait-0.0.9/lib/rspec/wait/target.rb:44:in `with_wait'
     # /usr/local/bundle/gems/rspec-wait-0.0.9/lib/rspec/wait/target.rb:30:in `to'
     # ./spec/integrations/kill_word_spec.rb:12:in `block (2 levels) in <top (required)>'
     # ./spec/spec_helper.rb:19:in `block (2 levels) in <top (required)>'
[...]

PR #551, opened 2020-07-08
Preserve ZLE_KILL and ZLE_YANK flags

macdems:
This solves #363.

### DIFF
--- a/src/bind.zsh
+++ b/src/bind.zsh
@@ -51,8 +51,14 @@ _zsh_autosuggest_bind_widget() {
 	# correctly. $WIDGET cannot be trusted because other plugins call
 	# zle without the `-w` flag (e.g. `zle self-insert` instead of
 	# `zle self-insert -w`).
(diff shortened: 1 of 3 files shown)


vincentbernat:
I am running into an odd bug when applying this commit. Steps to reproduce:

1. Start a fresh shell
1. Type `echo hello world bye`
1. Type `Alt-Backspace` (this deletes bye and put it in the ring)
1. Type `Backspace` (this deletes the space)
1. Type `Alt-Backspace` (this deletes world and put it in the ring)
1. Type `Alt-b` (go at the beginning of `hello`)
1. Type `Ctrl-Y` (paste `world`)
1. Type `Alt-Y` (should replace `world` by `bye`)

What I get is:

```
echo hbyeello
```

Instead of:

```
echo byehello
```

ZSH 5.8. Do you have the same bug?

aculich:
@vincentbernat I added your suggestion as a test and opened a new PR #795 that builds on this one... hopefully they'll both be merged soon @macdems ?

PR #555, opened 2020-08-02
Ignore compiled files

druckdev:
When compiling before sourcing the plugin, a `zsh-autosuggestions.zsh.zwc` is created.
This can be annoying if zsh-autosuggestions is used as a git submodule in a dotfile repo for example, since the repository will then be marked as dirty.
This PR just adds a gitignore that ignores all files ending on `.zwc`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+# Ignore compiled files
+*.zwc


z0rc:
If you're compiling multiple zsh plugins it's sensible to add `*.zwc` to your global gitignore.

druckdev:
Oh wow! I did not know that was possible/did not expect this would solve my problem. Thanks!

I can still see it helping others eventually; what do you think? Otherwise this could be closed.

ericfreese:
I'm going to go ahead and close this for now but would consider it in the future if it comes up again.

druckdev:
Sounds reasonable. Thanks!

PR #556, opened 2020-08-08
Add official package for Debian and Ubuntu

FranklinYu:
OBS repository is provided as fallback, since the official package is not included in all supported releases.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -9,7 +9,8 @@
 
 | System  | Package |
 | ------------- | ------------- |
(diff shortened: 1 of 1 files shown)


FranklinYu:
Actually Fedora package is also available at https://apps.fedoraproject.org/packages/zsh-autosuggestions, but the entire Packages site is 503 right now. Please advise whether we should add it.

FranklinYu:
Friendly ping. This pull request is still valid.

PR #564, opened 2020-09-05
POSIX_IDENTIFIERS compatibility

kittokin:
zsh-autosuggestions plugin doesn't work when zsh has posix_identifiers option enabled.
when set, it keeps complaining about `"bad math expression: operator expected at "name..."`
but if you surround it with `{}` it stops and works again


kittokin:
Forgot to mention it work without that option being set

Issue #587, opened 2021-03-08
Installation guide is not suitable to install this plugin aside others


keskinsaf:
### Describe the bug

Installation script for [oh-my-zsh] suggests installing that plugin alone. Instead, it should be indicated that this plugin can be installed asides other plugins.

### To Reproduce

 1. Install some others plugins. For example, install [zsh-syntax-highlighting] following installation guide.
 2. Follow [INSTALL.md] to install this plugin.
 3. Check whether both plugins are working correctly or not.

### Expected behavior
Installation of this plugins shouldn't affect others. It should be installed alongside other plugins.

### Desktop
 - OS + distribution: macOS Big Sur (version: 11.2.1)
 - Zsh version: 5.8
 - Plugin version: v0.6.4

[oh-my-zsh]: https://ohmyz.sh/ "oh-my-zsh"
[zsh-syntax-highlighting]: https://github.com/zsh-users/zsh-syntax-highlighting "zsh-syntax-highlighting"
[INSTALL.md]: https://github.com/zsh-users/zsh-autosuggestions/blob/master/INSTALL.md "INSTALL.md"

PR #588, opened 2021-03-08
Update INSTALL.md

keskinsaf:
[fix] install this plugin aside other plugins

[#587]

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -39,7 +39,7 @@
 2. Add the plugin to the list of plugins for Oh My Zsh to load (inside `~/.zshrc`):
 
     ```sh
-    plugins=(zsh-autosuggestions)
+    plugins=( [plugins...] zsh-autosuggestions)
     ```
 
 3. Start a new terminal session.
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -39,7 +39,7 @@
 2. Add the plugin to the list of plugins for Oh My Zsh to load (inside `~/.zshrc`):
 
     ```sh
-    plugins=(zsh-autosuggestions)
+    plugins=( [plugins...] zsh-autosuggestions)
```
ericfreese, line 42: Can we write this in such a way that the `[plugins ...]` part can't be  confused for actual code? Could we do something like:
```
plugins+=(zsh-autosuggestions)
```

Or maybe this?

```zsh
plugins=(
  # Other plugins...
  zsh-autosuggestions
)
```

keskinsaf: I think second option is better, also more similar to other plugins.

I have updated pr accordingly

ericfreese:
Thank you!

Issue #364, opened 2018-08-07
A possibly needed workaround for <( ) process substitution


psprint:
Hello,
in branch `develop` there's currently use of `<( ... )` process substitution to do asynchronous calls: https://github.com/zsh-users/zsh-autosuggestions/blob/681ffc7b2891e97801a2104997ed8524fa2c4556/zsh-autosuggestions.zsh#L717-L725

I use the same technique in fast-syntax-highlighting, however, I add `command sleep 0` (could make it `command true`) after the `exec`: https://github.com/zdharma/fast-syntax-highlighting/blob/e06288a617b12e8038fdf5bc4bd40b6a25ef5332/fast-highlight#L917-L918

The reason for this additional call are severe problems with Ctrl-C that occur when using `{MYFD}<` construct, [as my investigation showed](http://www.zsh.org/mla/workers/2018/msg00966.html).

So in this issue I just want to mention that in case of any Ctrl-C problems, `command true` after exec should help.

ericfreese:
Thanks for this! I was actually having issues with Ctrl-C on develop but wasn't sure exactly why and hadn't yet put in the effort to figure out what was going on. I'm busy for the next few weeks but I'll take a closer look at this before cutting a release from develop.

ericfreese:
@psprint Did you ever figure out why exactly this was necessary? I'm really curious what's going on that makes ctrl-c stop working.

ericfreese:
I've found something interesting that you may have also found. The problem only occurs with the `monitor` option set (does not occur with it unset).

psprint:
Interesting. I was tracking this for many days and when workaround appeared I was too tired to continue, so I dont know why this problem occured

ericfreese:
Fair enough. I'll see what I can find...

ericfreese:
@psprint I've submitted a patch that fixes this issue upstream

http://www.zsh.org/mla/workers/2019/msg00252.html

ericfreese:
FYI @psprint it looks like this workaround doesn't work for zsh versions older than 5.0.8.

ericfreese:
Closing this. I've added the workaround with PR #417

psprint:
Great job!

Wt., 9.04.2019, 20:36 użytkownik Eric Freese <notifications@github.com>
napisał:

abcdw:
Can we use `command -v true` instead of `command true` here?

[...]

PR #612, opened 2021-06-03
Disable ^C async workaround for zsh version 5.8 and later

abcdw:
The bug is fixed in recent version of zsh, so the workaround is not needed anymore for fresh zsh:

https://github.com/zsh-users/zsh-autosuggestions/issues/364

### DIFF
--- a/zsh-autosuggestions.zsh
+++ b/zsh-autosuggestions.zsh
@@ -810,7 +810,8 @@ _zsh_autosuggest_async_request() {
 
 	# There's a weird bug here where ^C stops working unless we force a fork
 	# See https://github.com/zsh-users/zsh-autosuggestions/issues/364
-	command true
+	autoload -Uz is-at-least
+	is-at-least 5.8 || command true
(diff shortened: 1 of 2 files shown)


zsh-autosuggestions.zsh
```diff
@@ -810,7 +810,8 @@ _zsh_autosuggest_async_request() {
 
 	# There's a weird bug here where ^C stops working unless we force a fork
```
ericfreese, line 811: Thanks for the quick PR. Not a huge deal, but can you update this comment to read something like "There's a bug in older versions where ^C ..."?

ericfreese:
Just leaving a reminder for myself that I need to add 5.8 to the ZSH_VERSIONS file and push a new docker image

abcdw:
Sorry, I missed this email, when I rediscovered it the pr was already
merged.

abcdw:
Thank you for making a tag and for all your work!)

PR #621, opened 2021-07-14
Match multiple previous commands & more configuration

druckdev:
- Add `ZSH_AUTOSUGGEST_MATCH_PREV_MAX_CMDS` option: With this the until now hard-coded value of 200 can be configured.
-  Add `ZSH_AUTOSUGGEST_MATCH_NUM_PREV_CMDS` option: With this the number of preceding commands that have to match when using the `match_prev_cmd` strategy can be set (and can be multiple now).

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,8 @@ For example, setting `ZSH_AUTOSUGGEST_STRATEGY=(history completion)` will first 
 
 #### When `ZSH_AUTOSUGGEST_STRATEGY` contains `match_prev_cmd`:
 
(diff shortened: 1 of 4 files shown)


druckdev:
Wanted to create the PR already to get some feedback if what I am doing here is wanted.
I would look into writing some tests if that is the case.

I also wanted to ask what you think of an additional option that results in the suggestion being the command that has the *most* matching previous commands as fallback instead of *exact*. For the case when there is no matching command with the number of matching previous commands, as this could be the case on higher values for  `ZSH_AUTOSUGGEST_MATCH_NUM_PREV_CMDS`.

druckdev:
Oh, and also:
What do you think of the names of the two configuration options? I am a bit unsure if those are good names but have no better idea...

PR #156, opened 2016-05-11
Adjust plugin.zsh file to run on zsh 5.1 in mSYS2.

adamkruszewski:
I have fixed the missing brackets in plugin.zsh file. Now it works flawlessly under msys2 on Windows.
I have closed the other pull request as, being frank, I didn't know how to add a somthing from a different branch to it.

PR #640, opened 2021-11-20
Create a symlink instead of file

grigorii-horos:
`${0}` not compatible with some zsh plugin managers

### DIFF
--- a/zsh-autosuggestions.plugin.zsh
+++ b/zsh-autosuggestions.plugin.zsh
@@ -1,4 +1,0 @@
-0="${ZERO:-${${0:#$ZSH_ARGZERO}:-${(%):-%N}}}"
-0="${${(M)0:#/*}:-$PWD/$0}"
-
-source ${0:h}/zsh-autosuggestions.zsh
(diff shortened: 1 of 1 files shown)


ericfreese:
My understanding is that the sym link doesn't work in some environments like windows mSYS2 (see PR #156). It looks like zsh-syntax-highlighting may have an approach that works for both. Can you try implementing this here instead?

https://github.com/zsh-users/zsh-syntax-highlighting/blob/c7caf57ca805abd54f11f756fda6395dd4187f8a/zsh-syntax-highlighting.plugin.zsh

grigorii-horos:
> My understanding is that the sym link doesn't work in some environments like windows mSYS2

Oops… I didn't know.

Maybe we can use [this](https://github.com/zdharma-continuum/Zsh-100-Commits-Club/blob/master/Zsh-Plugin-Standard.adoc#zero-handling) for $0 handling?

ericfreese:
@grigorii-horos I've created a PR (#759) that takes the implementation from zsh-syntax-highlighting. I realize this PR is very old, but take a look if you're able. I'll likely merge it soon.

PR #654, opened 2022-02-06
Shallow cloning when installing locally

AlfredJKwack:
Create a shallow clones with a history truncated to 1 commit when installing locally. There's no point to pulling everything in. It's faster and saves resources.


AlfredJKwack:
After reading through [this article](https://github.blog/2020-12-21-get-up-to-speed-with-partial-clone-and-shallow-clone/) it seems indeed clear that for the size and nature of the repo the costs will outweigh the benefits. 
Closing with this comment.

PR #680, opened 2022-04-03
Fix 'bad math expression' bug

bbugyi200:
The problem seems to be that when the POSTDISPLAY envvar is not set, `$#POSTDISPLAY` sometimes evaluates to `0POSTDISPLAY` instead of `0` as is desired.

Using curly braces to disambiguate should mitigate this issue. This should also be a low-risk change IMO.


bashenk:
I've noticed extremely similar behavior, at one time or another, myself. It might be a good idea to consider doing this for each reference of `$#POSTDISPLAY` where it could theoretically be unset, since it does show up elsewhere (`src/widgets.zsh`, `src/highlight.zsh`, and the `_zsh_autosuggest_accept` function in the same file). If no one gets around to it I can fork and submit a PR at a later date.

PR #694, opened 2022-06-26
Don't need to clone full history when install

haunt98:
Use `--depth 1` to skip clone full history when install


z0rc:
I don't agree. This isn't recommended, see discussion at https://github.com/zsh-users/zsh-syntax-highlighting/pull/855 on reasons why.

Issue #678, opened 2022-04-03
Conflict with search history with text already entered at the prompt


tapyu:
### Describe the bug
Dears,

I am using the following snippet from a [stack exchange answer](https://unix.stackexchange.com/a/97844/480687):

```
autoload -U history-search-end
zle -N history-beginning-search-backward-end history-search-end
zle -N history-beginning-search-forward-end history-search-end
bindkey "^[[A" history-beginning-search-backward-end
bindkey "^[[B" history-beginning-search-forward-end
```

This allows me to search through my history from the already entered text at the prompt. However, the `zsh-autosuggestions` plugin has conflicts with this snippet, which obligates me to press uparrow twice in order to search through the history

### To Reproduce
1. Put this snippet in the ~/.zshrc
1. Check that you can search from the text already entered at the prompt pressing uparrow (just once per command)
1. Install the `zsh-autosuggestions` plugin
1. Notice that now you need to press the uparrow twice per command

### Expected behavior
I would like to use the `zsh-autosuggestions` plugin and perform this search history in a minimal configuration. At the moment, I am using the [joshskidmore/zsh-fzf-history-search](https://github.com/joshskidmore/zsh-fzf-history-search) plugin to make this task, but it is more desirable to use this snippet.

Anyone has an way out or can identify what is happening?

### Desktop
```
 OS: Ubuntu 20.04 focal
 Kernel: x86_64 Linux 5.13.0-35-generic
 Shell: zsh 5.8
 DE: GNOME 3.36.5
 WM: Mutter
 CPU: Intel Core i7-9750H @ 12x 4.5GHz [44.0°C]
 GPU: Intel Corporation UHD Graphics 630 (Mobile)
NVIDIA Corporation GP107GLM [Quadro P620] (rev a1)
Plugin version: v0.7.0
```

### Additional context
- I am not using any framework to install plugins, I've been using [this approach](https://www.youtube.com/watch?v=bTLYiNvRIVI), which works elegantly to all my plugins but this one...
- Notice that this simpler snippet (it does not put the cursor at the end of the searching)
```
bindkey "^[[A" history-beginning-search-backward
bindkey "^[[B" history-beginning-search-forward
```
works properly with the `zsh-autosuggestions` plugin. I think that the conflict is related to `history-search-end` load of the `zsh`.

hongwen000:
Same problem, I had trouble browsing history using up arrow key, twice pressing is needed, and the anoying terminal bell always rings. 

[...]

Issue #619, opened 2021-07-01
async mode breaks the history-search-end zle widget


doy:
### Describe the bug
async mode was recently enabled by default, and when used, `history-search-end` doesn't work properly. the exact way it fails is a bit inconsistent - it feels like there is a race condition or something. most commonly you need to press the key it is bound to multiple times to get it to do anything, and eventually it stops searching through history at all.

### To Reproduce

```
# .zshrc
export HISTFILE=$HOME/.zsh_history
export HISTSIZE=1000
export SAVEHIST=1000

autoload history-search-end
zle -N history-beginning-search-backward-end history-search-end
zle -N history-beginning-search-forward-end history-search-end
bindkey "^[OA" history-beginning-search-backward-end # or "^[[A" depending on your terminal
bindkey "^[OB" history-beginning-search-forward-end  # or "^[[B" depending on your terminal

source ~/zsh-autosuggestions/zsh-autosuggestions.zsh
# unset ZSH_AUTOSUGGEST_USE_ASYNC
```

if you populate your history with a few commands that share a prefix, you should see it working properly if you uncomment the `unset` line, but break if you comment it out. for instance, run these commands in sequence:

```
echo foo
ls foo
echo bar
ls bar
echo baz
ls baz
echo quux
ls quux
```

and then at the prompt, type `echo` and press the up arrow key a couple times. this should cycle through `echo quux`, `echo baz`, `echo bar`, `echo foo` on each keypress, but if async mode is enabled, i get `echo quux` from the first up arrow press, then `echo baz`, then it stays on `echo baz`, and going further back through the history requires multiple up arrow presses. eventually if there are enough entries, something else happens where it stops searching any further backward at all.

### Expected behavior
the `history-search-end` widget should work the same way in async mode as it does with async mode disabled

### Desktop
 - OS + distribution: arch linux, also reproduced in a debian container
 - Zsh version: reproduced on 5.7.1 and 5.8
 - Plugin version: 0.7.0

ericfreese:
Hi, I think that this is a similar bug to #241. See https://github.com/zsh-users/zsh-autosuggestions/issues/241#issuecomment-375812886 and https://github.com/zsh-users/zsh-autosuggestions/pull/319

Try adding those widgets to your `ZSH_AUTOSUGGEST_CLEAR_WIDGETS` via something like this in your zshrc:

```
[...]

PR #706, opened 2022-09-10
Add history-beginning-search-forward-end and history-beginning-search-backward-end to ZSH_AUTOSUGGEST_CLEAR_WIDGETS

pablospe:
Fixes issues #678. Based on modifications suggested here https://github.com/zsh-users/zsh-autosuggestions/issues/619#issuecomment-904193190. @artudytu @ericfreese


hongwen000:
I think this solution works, when will it be merged?

ericfreese:
Thanks. I have cherry-picked the commit as 23f2943 with some slight modifications to also update the source file at `src/config.zsh`

PR #713, opened 2022-11-02
Fix plugin for shells with `no_unset` shell option (aka `set -u`) set

ntninja:


ShayanAhmad (approved):
looks good to me.

REALERvolker1:
Just going to comment on this: `set -u` is useful in interactive contexts, and it would be nice to have it in autosuggestions.

z0rc:
`set -u` isn't useful in any way in interactive shell, see discussion at https://github.com/romkatv/powerlevel10k/issues/2652.

LucasLarson:
> `set -u` isn't useful in any way in interactive shell

is incorrect and the [latest message](https://github.com/romkatv/powerlevel10k/issues/2652#issuecomment-2070935444) in [romkatv/powerlevel10k #2652](https://github.com/romkatv/powerlevel10k/issues/2652) is incorrect. When you source a function like this:
```zsh
$ cat ./foo && . ./foo
break_nounset() {
  printf 'foo%s\n' "${this_variable_does_not_exist}"
}
```
it returns different results based on user settings
```zsh
$ set -u && break_nounset
break_nounset:1: this_variable_does_not_exist: parameter not set
$ set +u && break_nounset
foo
```

Having `set -u` in an interactive shell is invaluable while dogfooding works in progress.

LucasLarson:
> I must emphasize, zsh-autosuggestions is designed to run in interactive session, it's expected to have sane defaults

It appears that much of the codebase does allow for a user to work with it interactively, even if `set -o nounset`/`set -u`/`setopt nounset` is activated. In 2019, for example, the codebase was modified so that users with `set -o sh_word_split`/`set -y`/`setopt sh_word_split` in their interactive shell can keep easy access ([`4cd210b70d`](https://github.com/zsh-users/zsh-autosuggestions/commit/4cd210b70d20e24946d68a4957cb4a90ff97cc44)).

I see plenty of benefit in publishing changes that would allow this plugin to degrade gracefully and work for users coming with various options set.

What’s the benefit of preventing this merge?

JamesWidman:
+1 to @LucasLarson; `no_unset` is useful in interactive mode, and not just because it catches errors in scripts: a variable name can be typo'd (or erroneously unset) on the command prompt just the same as a variable in a script, and `no_unset` will prevent such an erroneous command on the prompt from being run. That's a very good thing if the erroneous command would produce irreversible effects or just consume a lot of time!

and for people who use `unset` in interactive mode, this PR does no harm.

there's now at least 4 different users who want this; is there any reason not to merge?

ed9w2in6:
Regarding no_unset (set -u), it is addressed at https://github.com/zsh-users/zsh-autosuggestions/pull/811, which is merged and release with v0.7.1.

Consider if this is still useful e.g. is `ksh_arrays` solved too?

Or at least remove the part with regarding to unset, then it can possible be merged if ksh_arrays is wanted by people.

PR #723, opened 2023-02-15
Create .gitignore for *.zwc

migimigi:
Created a .gitignore file that ignores the zsh word code (*.zwc) compiled with the zcompile command for zsh scripts.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+# zsh word code files
+*.zwc


graelo:
Thanks, yes I'd like that too.

For instance, <https://github.com/zsh-users/zsh-syntax-highlighting.git> has this in their `.gitignore`

```text
*.zwc*
```

ericfreese:
Thanks!

PR #724, opened 2023-02-17
Don't discard $WIDGETSTYLE

marlonrichert:
Calling a completion widget without passing the -w flag causes $WIDGETSTYLE to be unset. This breaks logic in _oldlist, _list_files and some plugins.


marlonrichert:
It appears that this fixes completion widgets, but in turn break normal widgets. This needs to be solved in a different way.

PR #741, opened 2023-05-20
Adding support for ZSH_AUTOSUGGEST_BUFFER_MIN_SIZE to disable suggestions on small buffers

msohaill:
As the title states, sometimes suggestions on small buffer can get annoying especially for 1-3 characters. These changes add support for a buffer min size. Added/updated tests accordingly.


msohaill:
Very odd, I cannot seem to figure out the reason for the `yank-pop` test failures. It seems something very specific or OS/system related, as running the tests locally with the provided docker image is always successful for me. I'll leave this open

Issue #737, opened 2023-05-07
`zle` widgets starting with `_` are ignord in `ZSH_AUTOSUGGEST_CLEAR_WIDGETS`


happenslol:
### Describe the bug
Any widget that has a name starting with an underscore will not cause the prompt to be cleared. I'm not sure if this is intended behavior, but it took me quite a while to figure out why some widgets were not clearing my prompt.

### To Reproduce

```zsh
function _test() {}
zle -N test-widget _test
zle -N _test-widget _test

bindkey "^a" test-widget
bindkey "^b" _test-widget

ZSH_AUTOSUGGEST_CLEAR_WIDGETS+=('test-widget')
ZSH_AUTOSUGGEST_CLEAR_WIDGETS+=('_test-widget')
```

Trigger any suggestion. On pressing `^a`, the suggestion is cleared, on `^b` it is not.

### Expected behavior
Any widget name should be able to work. Other zsh plugins I'm using add widgets that start with `_`, which means I currently have to rebind them to let them clear the suggestions.

### Desktop
 - OS + distribution: `NixOS 23.05`
 - Zsh version: `5.9`
 - Plugin version: `v0.7.0`

PR #749, opened 2023-07-05
Allow wrapping widgets whose name start with _ (underscore)

toadjaune:
Hi folks,

When creating a zle widget whose name starts with `_`, it is ignored by zsh-autosuggestions, leading to various situations where the suggestion should be refreshed, but isn't.

Typically, other plugins often create such widgets, making them incompatible with zsh-autosuggestions.

This PR changes this.

Preliminary testing seems to indicate everything works fine on my machine, and before going through more extensive testing and/or writing tests, I wanted to ask if there's a specific reason to forbid wrapping such widgets.

It _looks like_ the current presence check is not a mistake, but a voluntary design, but I can't understand why. As far as I can tell, the only reason to ban such widget names, is to prevent wrapping an already-wrapped widget.
Which the proposed change still does.

https://github.com/zsh-users/zsh-autosuggestions/issues/737 is a manifestation of this issue, and provides a nice minimal reproducing example (although the problem is bigger than stated there, as it affects all types of widget wrapping)

If we go through with this change, it would likely require a `0.8.0` tag, as suddenly starting to wrap previously-unwrapped widgets may break unexpectedly on specific edge cases. It is however possible for users with such a setup to reproduce the current behavior with `ZSH_AUTOSUGGEST_IGNORE_WIDGETS`, so, it's not a blocking change.

Please let me know if I've missed the rationale behind the current design !

### DIFF
--- a/src/bind.zsh
+++ b/src/bind.zsh
@@ -68,7 +68,7 @@ _zsh_autosuggest_bind_widgets() {
 
 	ignore_widgets=(
 		.\*
-		_\*
+		_zsh_autosuggest_\*
 		${_ZSH_AUTOSUGGEST_BUILTIN_ACTIONS/#/autosuggest-}
 		$ZSH_AUTOSUGGEST_ORIGINAL_WIDGET_PREFIX\*
 		$ZSH_AUTOSUGGEST_IGNORE_WIDGETS
--- a/zsh-autosuggestions.zsh
+++ b/zsh-autosuggestions.zsh
@@ -198,7 +198,7 @@ _zsh_autosuggest_bind_widgets() {
 
 	ignore_widgets=(
 		.\*
-		_\*
+		_zsh_autosuggest_\*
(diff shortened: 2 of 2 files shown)


norman-zon:
Any chance of having this merged an released soon?

norman-zon:
bump?

toadjaune:
Hi folks, I originally opened this PR to fix compatibility issues with zsh-syntax-highlighting ; which is no longer the case because they switched to using a zle-line-pre-redraw hook instead of wrapping widgets on newer zsh versions.

Still, it's probably still a good idea to merge this, as a widget starting with an underscore is a perfectly valid name, and there's no apparent reason not to accept wrapping such a widget if the user asks for it.

toadjaune:
I'm unsure who to ping, maybe @ericfreese , since you're the last person to have merged something on master ?

PR #755, opened 2023-08-17
Update README.md

rweir:
Clarify where to get `zpty` from.

Info from https://apple.stackexchange.com/a/416099 and https://zsh.sourceforge.io/releases.html .


ericfreese:
Thanks!

PR #758, opened 2023-09-03
Update INSTALL link for Mac OS - homebrew moved under z/

mjperrone:
See https://github.com/Homebrew/homebrew-core/commit/442f9cc511ce6dfe75b96b2c83749d90dde914d2

Also, it might be nice to include `brew install zsh-autosuggestions` there as the package definition doesn't help someone with how to actually _install_ it unless they know how homebrew works. Also, if you aren't aware of homebrew's `Caveats`, it might be easy to miss

```
==> Caveats
To activate the autosuggestions, add the following at the end of your .zshrc:

  source /opt/homebrew/share/zsh-autosuggestions/zsh-autosuggestions.zsh

You will also need to restart your terminal for this change to take effect.
```

So I would also suggest adding that to your install documentation page.


ericfreese:
Thanks!

PR #790, opened 2024-05-22
Fix colors not being reset on (partial) accept w/ recent ZSH (fixes #789)

### LINKED ISSUE #789: Coloring doesn't change on auto suggestions
### Describe the bug

After updating to Ubuntu 24.04, I don't get color changes anymore on auto-suggestions. See screenshot. 

### To Reproduce

Sorry haven't found a way to reproduce this. But setting it up from scrats on a new VM it's working like expected. So it's probably something local. :( 

### Expected behavior
I would expect the coloring to change, to the primary color as before the update.

### Screenshots
zsh-autosuggestions-coloring

### Desktop
I run following versions: 

- Ubuntu 24.04 (Budgie Edition)
- kitty 0.32.2
- zsh 5.9
- oh-my-zsh (commit: b1c5315)
- zsh-autosuggestion (commit: c3d4e57)

ElementW:
The ZSH manual describes `region_highlight` as being an array in https://zsh.sourceforge.io/Doc/Release/Zsh-Line-Editor.html#Character-Highlighting, therefore the previous strategy of removing as many characters as the last suggestion is *not* the way to do it, explaining why it broke on recent ZSH versions.

Replace this logic with a simple last-element delete. Keeps the `_ZSH_AUTOSUGGEST_LAST_HIGHLIGHT` variable intact since there's no downside in tracking its content, as it still used as a marker for whether a suggestion highlight was applied.


tomasnorre:
I can confirm that this fix solves the problem I have in #789

ErwanGarreau:
I had the exact same problem (the zsh-autosuggestions proposed color was not changing back although it was accepted).
This commit also fixed it for me. Thank you !

rylwin:
This commit also fixes it for me, too. I started experiencing this issue after upgrading to Ubuntu 24.04 (from 22.04) 

```
$  zsh --version
zsh 5.9 (x86_64-ubuntu-linux-gnu)
```

dolphin-cat:
also fixes it for me, #820

enoryw:
What's preventing this from getting merged?

rieje:
Fixed for me too. Can this be merged? It's nearly a year and colors being broken for an autosuggestion plugin that relies on colors is kind of essential...

rbjorklin:
I was having the same problem until I started using the code from this PR.

@ericfreese any chance you have a moment to review and merge?

PR #794, opened 2024-07-14
Fix yank-pop and add test for cursor positioning

### LINKED ISSUE #363: Autosuggest eats buffer by ctrl+w

aculich:
Based on suggestion in PR #551 that notes a problem with yank-pop cursor positioning.

Fixed by moving `yank` to:

`ZSH_AUTOSUGGEST_ZLE_YANKBEFORE_WIDGETS`

instead of:

`ZSH_AUTOSUGGEST_ZLE_YANK_WIDGETS`

Also confirmed that this works to close #363 once merged in.


aculich:
Oops, this was supposed to be for develop, rather than master :(

PR #826, opened 2025-04-17
allow running with set -u (no unset var)

ed9w2in6:
I have been using this for a while.
It makes me feel safer by turning on set -u by default, which abort on unset var

This line seems to be the only one tripping set -u so this is a very simple change.


ed9w2in6:
the problem this pull request attempts to solve is already solved at https://github.com/zsh-users/zsh-autosuggestions/pull/811, great.

